PR #1, opened 2026-09-16
Constrained reconstruction: preregister problem and implement compact candidate

Liang-techh:
Implements the first nonzero compact axisymmetric candidate and a separately restricted two-coefficient force. Preregisters domain, viscosity, energy normalization, parameter bounds, held-out sampling and acceptance thresholds. Velocity uses an axis-regular streamfunction plus swirl; candidate parameters and initialization diagnostics are saved under artifacts/constrained.

Validation: 5 focused tests pass (independent finite-difference force curl and velocity divergence, axis/support/core signs, normalization/scaling and artifact round-trip). Reference energy at quadrature orders 24/48/96: 1.0000089576 / 1.0000000018 / 1.0. git diff --check passed.

CR001 and CR002 delivered. CR003 derivative interfaces, compatible pressure, optimization, independent PDE validation and structural proofs remain pending. The initial zero pressure is not claimed compatible; the candidate is not an exact reproduction or a validated NS solution.


Liang-techh:
本轮代码、coupled_joint 候选和验证已上传。用户已把优先目标更新为获得与 OpenAI 可视化对应的可计算 u(x,y,z,t), v(x,y,z,t), w(x,y,z,t)。

定时 agent 请从 **#15** 和 `docs/SCHEDULED_AGENT_TASKS.md` 接手，按 VIS001–VIS006 逐项认领、交付并标记完成。旧 SCH 研究队列已延后，避免阻塞速度 API 和实际可视化交付。现有 PDE 失败结果保留，视觉对应仍待验证。

Liang-techh:
## Repository-wide NS routing update: advance from ST006

The retained ST006 baseline on `main` must now be treated as the repository-level PDE benchmark by **all** active NS agents, not only the Kokuno lane.

Independent replay:
- momentum max: **0.1082289305112118**
- seed `9172801`, 4096 held-out Cartesian points, six fixed times, finest spatial step `0.005`
- original target remains `1e-3`
- divergence gate still fails
- `pde_validated=false`

Routing consequence:
- If your candidate/result is directly comparable to this protocol and is worse than ST006, do **not** present it as repository-level PDE progress or promote it over ST006. It may remain useful as structure/capacity/source evidence.
- If compatible with your representation, prefer warm-starting / transferring from ST006 rather than restarting from O(1)-O(10) residual fields.
- If not compatible (e.g. source-faithful Kokuno mathematics), keep the route separate but explicitly use ST006 as the numerical benchmark to beat before PDE promotion.
- Do not import ST006 coefficients as source truth, reuse validation data for fitting, change forcing, or relax the fixed `1e-3` gate.
- Every future residual report must state whether it is directly comparable with ST006.

Coordination PR #290 now writes this rule into `PROJECT_GOAL.md`, `CURRENT_CHECKPOINT.md`, `AGENT_TASKS.md`, and `SCHEDULED_AGENT_TASKS.md`.

PR #10, opened 2026-09-16
CR003: probe minimal inner-annulus swirl capacity

Liang-techh:
Constrained Agent 7 minimal follow-up, stacked on the current constrained development head after checking open PRs #1–#8 and the latest quintic/frozen-collar diagnosis. This does not modify Agent 1's production candidate or optimizer.

### task_id
`CR003-INNER-ANNULUS-CAPACITY-002`

### Why this increment
The selected `quintic_swirl` family has an exact representation blind region: every current correction contains `parent.outer_basis`, which is identically zero for `r^2<=.125`. The recorded point `(r,z,t)=(.32,.885714..., .75)` lies inside that region and retains a refined residual norm about `1.15116`, with dominant azimuthal component about `-1.02161`. Increasing the old temporal polynomial degree cannot change the field or its coefficient Jacobian there.

### Prototype
Adds two diagnostic, not-production, axisymmetric pure-swirl directions `(-y,x,0) a(r^2,z^2)`:
- smooth inner guard exactly zero for `r<=.1` and fully on by `r>=sqrt(.02)≈.1414`;
- compact outer support through the repository cutoff;
- one broad mode and one localizer centered near the frozen collar;
- optional numerical angular-moment cancellation by subtracting a fixed multiple of the existing `parent.outer_basis`.

That subtraction does not erase the new inner-annulus reach because the old outer basis is exactly zero there. It also leaves the current 21 core probes untouched: their maximum radius is `0.08660 < .1`.

### Basis-growth evidence
On seven deterministic inner-annulus probe points, flattened evaluation rank grows from 1 to 2 when the second mode is added. After column normalization the two-mode condition number is `3.670851...`. At the frozen point the two new azimuthal components are about `0.32` and `0.31854`, while the current quintic correction basis is exactly zero on the full fourth-order residual stencil used at `h=.005`.

The focused regression also checks numerical divergence of both raw pure-swirl modes and selected-order angular-moment cancellation against the current parent.

### Truth boundary
This is a support/rank capacity result, **not** a PDE-residual improvement claim. Pressure, restricted force, parent candidate, thresholds and validation samples are unchanged. No free forcing and no `u->0` construction are used. A production candidate still needs bounded nonlinear fitting and independent held-out PDE validation before any acceptance claim.

### Files
- `src/openai_ns_reconstruction/constrained_inner_annulus_capacity.py`
- `tests/test_constrained_inner_annulus_capacity.py`
- `artifacts/constrained/inner_annulus_capacity.json`

### Remaining handoff
Agent 1 should test this or an equivalent tiny guarded inner-annulus swirl block before adding more old-family time degree. If the dominant residual rotates into radial/axial channels after that, the next minimal extension should be poloidal/pressure rather than more swirl modes.


Liang-techh:
Closing this capacity prototype because the active constrained base advanced while the branch was being prepared: `codex/cr001-constraints@1cf666a` now already implements the same guarded inner-collar pure-swirl extension as production `inner_swirl` (six rings / 30 new coefficients) and reports the frozen-point residual reduction. Keeping #10 open would duplicate Agent 1's now-landed lane. Agent 7 is pivoting to a non-overlapping residual-channel obstruction/sensitivity audit on the new result.

PR #12, opened 2026-09-16
CR003: diagnose axial channel capacity after inner swirl

Liang-techh:
Constrained Agent 7 follow-up on the latest active constrained head. I first prepared an inner-annulus swirl prototype, but closed PR #10 as soon as Agent 1's branch independently landed the same guarded inner-swirl idea. This PR is the non-overlapping pivot: residual-channel sensitivity after that production extension.

### task_id
`CR003-AXIAL-CHANNEL-OBSTRUCTION-003`

### Motivation
The production `inner_swirl` extension successfully attacks the former frozen collar, but its 64x64 cylindrical diagnosis moves the late-time peak into an axial-dominated channel near `(r,z)=(1.0154,.68794)`, with axial residual about `1.15338`. The subsequent bounded pressure refits remain O(1): `inner_swirl_pressure` has finest-step final-time sampled max `1.24417`, and the newer local-pressure expansion improves only to about `1.24361`. The pressure-independent circulation audit likewise reports an estimated lower bound about `.35715` for fixed velocity/force. This makes another pure-swirl or pressure-only growth cycle a poor next capacity experiment.

### Minimal diagnostic
Adds two fixed probe directions without modifying the production candidate:
- a compact axisymmetric **pure-swirl** mutation localized near the observed peak;
- a compact divergence-free **poloidal curl** mutation `curl((-y q,x q,0))` localized near the same region.

Pressure and the preregistered restricted force remain frozen. The tests use the repository's independent finite-difference residual operator, not the training loss.

The key channel regression verifies that a finite pure-swirl mutation changes radial/azimuthal momentum while its axial-residual change stays below `1e-10` on the selected peak-neighborhood points. A centered `±1e-3` coefficient probe verifies that one poloidal curl direction has axial-residual sensitivity above `1e-2`. Another test verifies finite-difference divergence below `2e-6` for the diagnostic poloidal curl on 24 seeded interior points and exact zero at selected exterior support probes.

### Actual verification
Exact-head Actions run `35127663344` completed **success**. The `constrained-integration` job ran `python -m pytest -q -W error tests/test_constrained_*.py` and reported **49 passed in 6.10s**. Coordinate, velocity, forcing and import/CLI jobs also passed; the optional legacy full matrix was skipped by workflow design.

### Interpretation boundary
This is a capacity/obstruction diagnostic, not a fitted candidate and not evidence of a smaller held-out PDE residual. It does not change forcing, pressure coefficients, thresholds, normalization, or candidate selection. No residual-defined force and no `u->0` construction are used. The intended handoff is only: test a **small bounded poloidal block jointly with pressure** before spending more budget on pure-swirl or pressure-only basis growth.

### Files
- `src/openai_ns_reconstruction/constrained_channel_capacity.py`
- `tests/test_constrained_channel_capacity.py`
- `artifacts/constrained/axial_channel_capacity.json`

No PDE acceptance, paper-exact reconstruction, or blow-up claim is made.


Liang-techh:
Agent 1 handoff pickup: I used this axial-channel result to open PR #13 with a bounded four-parameter compact poloidal correction primitive (`curl((-y q, x q, 0))`), analytic first/time derivatives, support handling, and serialization. I did not duplicate this PR's obstruction probe and did not fit pressure/PDE residual. Production joint poloidal+pressure fitting remains a later integration step after #13 is replayed onto the active constrained branch.

PR #17, opened 2026-09-16
CR009: audit coupled-candidate refinement plateau independently

Liang-techh:
Constrained Agent 3 minimal CR006/CR009 increment, replayed onto the latest active constrained head `b4f9d1d07753f97fa904656a2976b123c4944351` after the base advanced during the run. The stale first attempt was closed rather than left as a duplicate.

### task_id
`CR009-COUPLED-REFINEMENT-AUDIT-002`

The active branch already contains the independent fourth-order/second-order PDE validator and a three-level held-out validation for the current `coupled_joint` candidate. PR #5 calibrates the verifier on a nonlinear manufactured solution, while PR #8 adds an off-grid sampler. This PR adds neither. It adds the missing interpretation layer for the **current exact candidate's existing three-level held-out refinement**, explicitly separating an O(1) momentum plateau from finite-difference divergence contraction.

### Increment
Adds `constrained_refinement_audit.py`, which consumes only independent validation rows plus preregistered thresholds. It never imports the optimizer/training loss and never reevaluates or fits the candidate. The audit requires at least three common derivative levels at every validation time and fail-closes on malformed/incomplete rows.

A momentum metric is marked plateau-like only when, between the two finest levels, it changes by <=10% while remaining >=100x above the preregistered threshold. Divergence is classified separately as still discretization-sensitive when the fine/medium ratio is <=0.5. These are conservative diagnostics, not a proof of an h->0 limit.

### Current coupled_joint evidence
Source: `artifacts/constrained/coupled_joint/validation.json`, validation seed `914027`, 4096 held-out points, derivative steps `0.02, 0.01, 0.005`.

At the finest step across the six preregistered times:
- momentum sampled max: `0.604225..1.002218` = `604..1002x` the `1e-3` threshold;
- momentum L2: `1.169648..2.221532` = `1169..2222x` threshold;
- medium-to-finest momentum max changes are only `0.13%..7.41%` (all <=10%);
- divergence fine/medium ratios are `0.063..0.270` for sampled max and `0.064..0.316` for L2.

Thus all six times are classified `momentum_mismatch_not_explained_by_current_fd_refinement`: the O(1) momentum failure is not behaving like the simultaneously contracting finite-difference divergence error at these levels. This is evidence of a candidate/PDE mismatch plateau at the current refinement levels, not a rigorous continuum lower bound.

### Actual checks before upload
- isolated focused test: `PYTHONPATH=. python -m pytest -q -W error test_refinement_audit.py` -> `3 passed in 0.05s`
- `python -W error -m py_compile ...` -> exit 0

Repository exact-head CI will additionally run the new test against the checked-in `coupled_joint/validation.json`.

### Truth boundary / remaining limitation
No threshold is changed. No training sample or loss is reused. No forcing/pressure/velocity parameter changes. This PR does not make `coupled_joint` pass PDE or divergence acceptance, does not prove the residual floor rigorously, and makes no paper-exact or blow-up claim. CR009 perturbation/generalization across parameter/initialization changes remains open.


Liang-techh:
Exact-head CI for `f91adbe65e183803f731b25d61ad76980ffd5098` is green: Actions run `35130131324` completed success. `constrained-integration` ran `python -m pytest -q -W error tests/test_constrained_*.py` and reported **51 passed in 5.49s**; coordinate, velocity, forcing, and import/CLI jobs also succeeded, with the optional full matrix skipped by workflow design. This remains a CR009 refinement-interpretation check only: the coupled candidate still fails the preregistered PDE thresholds, and no continuum lower bound or acceptance claim is made.

PR #50, opened 2026-09-16
CR006: cross-check public velocity with alternate PDE stencil

Liang-techh:
Constrained Agent 3 minimal CR006/CR009 increment from the active constrained integration head `codex/cr001-constraints@f9d700b40f1335d7287ad5a8f7c7c2b2b452dc55`, after rereading main `AGENTS.md`, `README.md`, `docs/PROJECT_GOAL.md`, `docs/CURRENT_CHECKPOINT.md`, `docs/AGENT_TASKS.md`, then checking current open PRs, recent commits, and Actions.

### task_id
`CR006-ALT-PUBLIC-OPERATOR-005`

### Why this is not duplicate work
- PR #5 calibrates the existing independent fourth-order-space / second-order-time PDE operator.
- PR #17 interprets the current three-level residual plateau.
- PR #27 audits visualization-grid interpolation against direct public velocity calls.
- PR #34 audits bounded velocity-parameter perturbations.
- PR #42 routes the existing fourth-order PDE validator through the public `VelocityField.at_points(...)` interface.

PR #42 explicitly leaves an alternate-operator cross-check as an unowned follow-up. This PR fills only that gap: it independently codes a **second-order Cartesian spatial residual/divergence operator**, then compares it against the existing fourth-order independent operator on exactly the same public velocity artifact. It does not add another optimizer, candidate family, pressure basis, force model, or visual metric.

### Increment
Adds `constrained_public_velocity_alt_pde.py`:
- obtains every `[u,v,w]` sample through `field.at_points(...)` only;
- independently implements centered second-order spatial velocity derivatives, Laplacian and pressure gradient, plus second-order time differentiation with one-sided endpoint stencils;
- leaves pressure and the preregistered restricted force explicit;
- compares alternate-operator and existing fourth-order residual/divergence arrays on the same points, times and at least three derivative steps;
- reports sampled max/L2 metrics, operator-to-operator residual/divergence deltas, and ratios to the **unchanged** preregistered PDE/divergence thresholds;
- fail-closes on malformed public velocity, malformed pressure/force output, invalid bounds/steps, or missing/nonpositive thresholds;
- always keeps `pde_validated=false`; cross-operator agreement is evidence against one stencil-specific implementation artifact, not a PDE acceptance claim.

### Actual isolated verification before upload
Using the exact new module/test logic against a fixture containing the repository reference fourth-order operator:
[...]

PR #27, opened 2026-09-16
CR009: audit visualization-grid sampling against direct velocity

Liang-techh:
Constrained Agent 3 minimal CR009 increment, branched from the active integration head `30f5cbe7febc7a0c45d65df4f4c6ca7a02a37cb7` after reading the current function-first checkpoint, open PRs, and existing CR006/CR009 work.

### task_id
`CR009-VISGRID-001`

### Why this is not duplicate work
PR #8 supplies independent Sobol sampling for residual coverage. PR #17 interprets the existing three-level PDE refinement of `coupled_joint`. This PR does neither. It checks a different failure mode required by the current velocity-first goal: whether a finite visualization grid can create or hide structure relative to direct off-grid calls to the frozen public velocity API.

### Increment
Adds `constrained_velocity_grid_consistency.py`:
- takes only a `VelocityField`-like public interface (`grid` and `at_points`), never optimizer/training internals;
- samples one frozen set of held-out physical points per time and reuses it at all grid levels;
- trilinearly interpolates sampled vector grids back to those off-grid points;
- reports max vector error, RMS vector error and RMS error normalized by direct velocity RMS;
- requires at least three strictly increasing grid levels and fail-closes on malformed/nonfinite data;
- includes a one-command packaged-candidate report via `python -m openai_ns_reconstruction.constrained_velocity_grid_consistency`.

The focused tests include exact affine-vector interpolation, a smooth nonlinear field showing three-level refinement, malformed-grid rejection, and a real `VelocityField()` integration that loads the packaged candidate through the user-facing artifact path. The constrained CI lane also prints the packaged-candidate numerical report so the sampling error is visible rather than inferred from a green test.

### Local isolated check before upload
Using the same audit/test code with a minimal local implementation of the public `VelocityField` contract (network is unavailable in the automation container):
`PYTHONPATH=. python -m pytest -q -W error tests/test_constrained_velocity_grid_consistency.py` -> `4 passed in 0.04s`.
The synthetic nonlinear refinement report gave coarse-to-fine RMS ratios `0.1104, 0.1212, ...` for the representative times and finest relative RMS around `2.7e-5..3.4e-5` for the shown first two times. These numbers calibrate the audit implementation only; they are not evidence about the repository candidate.

### Truth boundary
[...]

PR #58, opened 2026-09-16
CR009: separate temporal PDE refinement from spatial stencil

Liang-techh:
Constrained Agent 3 minimal CR006/CR009 increment from active integration head `47f465bcf91a730c9bd43d15469f2650bf90b360`, after rereading AGENTS/README/PROJECT_GOAL/CURRENT_CHECKPOINT/AGENT_TASKS and checking all current open PRs, recent commits and CI.

### task_id
`CR009-TEMPORAL-REFINEMENT-006`

- PR #17 interprets the existing three-level refinement where one `step` changes **space and time together**.
- PR #27 audits visualization-grid interpolation versus direct public velocity.
- PR #34 audits bounded candidate-parameter perturbations.
- PR #42 routes the fourth-order independent PDE validator through the public velocity API.
- PR #50 adds an alternate second-order **spatial** operator cross-check.

The remaining CR009 attribution gap is temporal differentiation: a residual plateau cannot be cleanly blamed on the candidate rather than the time stencil if spatial and temporal steps always change together. This PR holds the fourth-order Cartesian spatial stencil fixed and varies only the second-order time step.

### Increment
Adds `constrained_public_velocity_time_refinement.py`:
- obtains every `[u,v,w]` sample through `field.at_points(...)` only;
- computes the fourth-order spatial Jacobian/Laplacian and pressure gradient once at a declared fixed spatial step;
- varies only the second-order centered/one-sided time-difference step across at least three levels;
- reports sampled residual max/L2, component RMS, divergence max/L2, unchanged-threshold ratios, and adjacent momentum changes between time-step levels;
- supports endpoint one-sided stencils inside the preregistered `[0.25,0.75]` window;
- keeps pressure and forcing explicit and frozen;
- fail-closes on malformed public velocity, pressure/forcing, time levels, bounds, or thresholds;
- always records `pde_validated=false` and allows a visually stable field to remain a `visualization candidate only`.

### Regression design
1. Manufactured divergence-free field `u=(t^3 sin(y),0,0)` with exact fixed force. At interior and both time-window endpoints, the isolated temporal truncation error is second order, so steps `0.08/0.04/0.02` must reduce residual maxima by approximately `4x` per halving while the fixed spatial error is negligible.
2. A force-sign mutation leaves an O(1) residual plateau while the same time refinement is applied; this catches the false inference that solver/refinement activity by itself implies PDE agreement.
3. Public-interface-only behavior and malformed/two-level inputs fail closed.
4. A repository integration smoke loads the frozen packaged `VelocityField`, uses seed `914059`, keeps its existing pressure and restricted fitted force fixed, and runs the three preregistered temporal levels while spatial step stays fixed at `0.005`.

### Truth boundary
This increment does not modify candidate parameters, pressure coefficients, force coefficients, training data, validation thresholds, or visual targets. Temporal convergence is only numerical evidence about one discretization axis. Even a stable residual across temporal refinement is not a rigorous continuum lower bound; a finite sampled threshold crossing is not full PDE validation; visual stability does not imply PDE validity; and nothing here identifies the exact OpenAI field, proves paper exactness, singularity, or blow-up.

### Verification status
Exact-head CI is authoritative because the integration smoke needs the repository artifact tree. At PR creation time it has not completed, so no test/CI success is claimed prematurely.

acceptance: pending
merge_status: open


Liang-techh:
Exact-head verification update for `2e4f36785970b3d6d24ca9fe94f89d90a9a71e96`:

- GitHub Actions run `35158475228` -> **completed / success**.
- `constrained-integration` performed a fresh Ubuntu checkout, `python -m pip install -e '.[dev]'`, then `python -m pytest -q -W error tests/test_constrained_*.py` -> **72 passed in 5.54s**. This includes the four new temporal-refinement regressions.
- `slice-forcing`, `slice-velocity`, `slice-coordinates`, and `import-and-cli-help` -> **success**.
- optional legacy full matrix -> skipped by workflow design.

This verifies the time-step-only audit in the repository environment. It does not change or promote PDE/visual/paper claim state: `pde_validated=false` remains explicit; refinement evidence is not PDE acceptance.

PR #73, opened 2026-09-17
CR003: compose serializable Eq. 4.5 velocity candidate

Liang-techh:
## Task
`CR003-EQ45-VELOCITY-CANDIDATE-010`

## Base choice
I reread the requested `main` project instructions and checked current open PRs/commits/CI. `main@c6700851...` is still the migration baseline, while the active constrained integration branch has already safely integrated the source-correct q solver, Eq. (4.5) Cartesian evaluator, the divergence-preserving `Phi -> (v0,U)` adapter, and the bounded serializable `Phi/F` profile basis. This increment therefore starts from the current active integration head `codex/cr001-constraints@b03bde4ae9ad0ae9f123d1c88583b0af89a61f69` instead of duplicating those dependencies back onto stale `main`.

## Why this is not duplicate work
- the q solver and Cartesian Eq. (4.5) evaluator are already integrated;
- PR #64 already reaped the `Phi -> (v0,U)` adapter;
- PR #48 is already merged onto the active branch with the bounded serializable `Phi/F` basis;
- PR #57 owns the later physical-space divergence-preserving support taper;
- PR #65 owns a Python visualization smoke, not candidate composition.

The missing production step is a single frozen object that composes the integrated pieces and can be directly called, saved, loaded, and passed to downstream validators/visualizers.

## Increment
Adds `Eq45VelocityCandidate` in `constrained_eq45_candidate.py`.

The object:
- owns one bounded `Eq45CompactProfileBasis` plus governed autonomous `h` and finite time interval;
- maps basis jets through the integrated streamfunction adapter to coupled `(v0,U)` and uses the independent basis swirl `F`;
- feeds the resulting `(v0,F,U)` provider into the integrated vectorized Eq. (4.5) Cartesian backbone;
- exposes `velocity(points,time)`, `at_points(points,time)`, `velocity_xyz(x,y,z,time)`, and `__call__`;
- rejects time extrapolation beyond `[0.25,0.75]`;
- rejects an explicitly zero coefficient block, preserving the nontriviality boundary;
- provides deterministic JSON save/load and a canonical SHA256 over the candidate payload;
- checks source-classification and truth-boundary metadata on load so a stored artifact cannot silently promote hidden profiles/PDE/paper-exact claims.

A checked deterministic seed artifact is added at `artifacts/constrained/eq45_velocity_candidate_seed.json`.

## Status / truth boundary
[...]

PR #57, opened 2026-09-16
CR003: add divergence-preserving physical support taper

Liang-techh:
Constrained Agent 1 minimal CR003 representation increment from latest `main@c6700851c0330c0c92938d25204d07fbabf588f4`, after rereading README/AGENTS/PROJECT_GOAL/CURRENT_CHECKPOINT/AGENT_TASKS and checking current open PRs, recent commits, and the active Eq. (4.5) integration state.

### task_id
`CR003-EQ45-PHYSICAL-TAPER-008`

### Why this is not duplicate work
- PR #48 owns the bounded/serializable `Phi/F` similarity-profile basis.
- PR #32 owns the `Phi -> (v0,U)` streamfunction coupling.
- PR #56 already integrated the vectorized Eq. (4.5) Cartesian evaluator onto the active constrained branch.
- PR #54 identified a real support-semantics blocker: compact support in `(X,eta)` does not by itself imply the preregistered physical support `r<2, |z|<2`, and specifically the current #48 defaults can map outside that physical cylinder.

This PR implements one of #54's explicit integration remedies rather than duplicating the audit: a small **physical-space exterior connection that preserves axisymmetric incompressibility by tapering at the streamfunction level**, instead of multiplying Cartesian velocity by a scalar.

### Increment
Adds `AxisymmetricPhysicalTaper`:
- accepts a base axisymmetric streamfunction `psi` plus cylindrical `(u_r,u_theta,u_z)` values;
- uses a C4 physical taper `T(r^2,z)` with an exact inner plateau and exact zero at/outside declared `r<R`, `|z|<H` support;
- forms `psi_tilde=T*psi`, giving
  - `u_r_tilde = T*u_r - (psi/r) d_z T`
  - `u_z_tilde = T*u_z + 2 psi d_(r^2)T`
  so the poloidal field remains streamfunction-generated instead of acquiring the divergence error caused by naive Cartesian multiplication;
- multiplies the axisymmetric swirl channel by the same taper; pure axisymmetric swirl remains divergence-free;
- avoids `1/r` evaluation on the axis and fail-closes if a supplied streamfunction violates the axis-regular `psi=0` contract there;
- exposes schema-tagged save/load metadata and classifies the taper shape/plateau as `autonomous_design`;
- keeps `pde_validated=false`, `paper_exact=false`, and `openai_field_identified=false` explicit.

[...]

PR #75, opened 2026-09-17
CR006: audit reloaded Eq45 candidate divergence via public velocity

Liang-techh:
Constrained Agent 3 minimal CR006/CR009 increment, stacked only on PR #73's new serialized `Eq45VelocityCandidate` so it does not duplicate the candidate-composition lane.

### task_id
`CR006-EQ45-DIVERGENCE-ARTIFACT-008`

Existing Agent-3 PRs already cover the older packaged field's independent PDE operator/calibration, public-API residual, alternate spatial stencil, spatial-grid interpolation, parameter perturbations, temporal derivative refinement, and temporal snapshot interpolation. PR #73 is the first frozen/loadable Eq. (4.5) candidate artifact; it has composition tests but no independent physical-space finite-difference divergence audit after serialization/reload.

### Increment
Adds `constrained_eq45_artifact_divergence.py`:
- reloads the serialized Eq. (4.5) candidate artifact;
- obtains every velocity sample only through `field.at_points(points,time)`;
- independently finite-differences all 9 Cartesian first derivatives with a centered second-order operator;
- reports sampled divergence max, RMS, a dimensionless RMS normalized by the sampled velocity-gradient Frobenius RMS, and observed convergence order across at least three strictly decreasing spatial steps;
- uses fixed held-out off-grid points/times and never reads optimizer/profile derivative tensors;
- fail-closes on malformed/nonfinite velocity, inactive velocity, invalid samples, or fewer than three derivative levels;
- keeps `pde_validated=false`, `physical_support_validated=false`, and `full_momentum_residual_assessed=false` explicit.

A mutation wrapper adds `epsilon*x` to the public u-component, creating a known divergence floor without touching candidate internals. The regression requires the pristine reloaded candidate's divergence error to contract across three steps while the mutation remains near its known nonzero floor.

### Truth boundary
This increment checks sampled incompressibility only. The current Eq. (4.5) candidate still lacks the physical support connection from PR #57 and has no compatible pressure/forcing artifact wired for a full independent momentum-residual audit. Therefore a divergence-convergent result is not a full Navier--Stokes validation, visual correspondence claim, paper-exact reconstruction, OpenAI hidden-field identification, singularity, or blow-up evidence.

### Verification status
Exact-head GitHub Actions is authoritative because this stacked branch contains the repository's constrained workflow and the serialized seed artifact. CI result will be recorded after completion.

Base dependency: PR #73 head `39bdea136ef09cec5e1eed2391dedae8e1ef40e0`.
Current implementation head: `3ece400c3572961e520fcc98bbd20463f12fba64`.
acceptance: pending
merge_status: open


Liang-techh:
Exact-head verification complete on `3ece400c3572961e520fcc98bbd20463f12fba64`.

Actions run `35167262553`: **completed / success**. The constrained job used Ubuntu 24.04 + CPython 3.11.16, installed the editable package, and ran exactly:

`python -m pytest -q -W error tests/test_constrained_*.py`

Result: **83 passed in 3.68s**. Import/CLI, forcing and velocity slices also passed; coordinates completed successfully in the same run; legacy full matrix was intentionally skipped by workflow policy.

Numerical regression bounds certified by the new tests on the reloaded seed artifact, using held-out seed `914103`, 128 off-grid points/times and spatial steps `0.06 -> 0.03 -> 0.015`: divergence RMS and max both strictly decrease at every refinement, both observed RMS convergence orders are `> 1.5`, and normalized RMS decreases from coarse to fine. The independent mutation adds `epsilon*x` to `u` with `epsilon=0.125`; at the finest step its measured divergence RMS is constrained to `(0.105, 0.145)` and remains at least 80% of its coarse value, so the audit detects a non-contracting divergence floor rather than falsely reporting convergence.

Truth boundary unchanged: this is an independent incompressibility check only. `full_momentum_residual_assessed=false`, `physical_support_validated=false`, `pde_validated=false`, visual correspondence/paper-exact/OpenAI-field claims remain false. PR remains stacked on #73 and should be retargeted to `codex/cr001-constraints` after #73 lands.

PR #72, opened 2026-09-17
CR003: audit selected core separable profile rank

Liang-techh:
Constrained Agent 7 minimal basis-growth increment from active integration head `codex/cr001-constraints@689b51cfd71c47fabdea70ea0df87a4ca41d5add`.

### task_id
`CR003-PROFILE-SEPARABLE-RANK-010`

### Why this is not duplicate work
- #48 owns an actual bounded serializable Eq. (4.5) spatial basis implementation.
- #63 owns the +1 affine-in-time coefficient lift diagnostic.
- #47 owns generic caller-supplied nested residual+visual basis screening.
- #57 owns physical-space streamfunction taper/support.
- the function-first lane already selected `PaperCoreSeries(PaperCoreReference(sigma=.5), maxdegree=14, eta_nodes=257)` and explicitly asked agents not to repeat the degree/sigma sweep.

This PR instead audits the **actual selected nonlinear inner seed** for low-rank separability, including radial derivatives that matter to streamline/vorticity geometry. It does not build a competing basis or exterior taper.

### Increment
Adds `constrained_profile_separable_rank.py`.

For profile values and radial-X derivatives sampled as `(X,eta)` matrices, the diagnostic normalizes value and derivative blocks separately, stacks them, and uses one SVD basis for both. For requested ranks it reports value error, radial-derivative error, worst error, singular spectrum, numerical rank, and rank-truncation condition number. The derivative block prevents a value-only fit from hiding gradient modes that can alter vorticity/streamline structure.

The selected-core helper evaluates the current function-first seed on `X in [0.001,0.409]`, `eta in [-0.9,0.9]`, using 25x65 samples and ranks 1/2/3/4/6/8. `rank_rtol=1e-3` and the `1e-3` compression tolerance are autonomous diagnostic choices, not PDE or visual-acceptance thresholds.

### Selected inner-seed result
The checked calibration gives an effective rank-4 knee for all three stacked value/derivative channels at relative singular cutoff `1e-3`:
- `phi`: rank 4 worst relative error `1.3781e-3`, cond `1.40e2`; rank 6 `9.92e-6`, cond `1.38e4`.
- `F` (swirl profile): rank 4 `5.5828e-4`, cond `2.17e2`; rank 6 `3.19e-6`, cond `3.17e4`.
- `U` (axial profile): rank 4 `1.0272e-3`, cond `8.62e1`; rank 6 `4.25e-6`, cond `1.55e4`.

[...]

PR #63, opened 2026-09-16
CR003: compare static basis with one affine temporal lift

Liang-techh:
Constrained Agent 7 minimal expression-capacity increment, branched from the current active integration head `codex/cr001-constraints@47f465bcf91a730c9bd43d15469f2650bf90b360` after rereading README/AGENTS/PROJECT_GOAL/CURRENT_CHECKPOINT/AGENT_TASKS, checking all current open PRs, recent commits, and Agent 7's existing capacity work.

### task_id
`CR003-TEMPORAL-BASIS-LIFT-009`

### Why this is not duplicate work
- PR #39 owns generic visualization-feature sensitivity for a fixed parameterization.
- PR #47 owns bounded joint residual+visual screening across nested **spatial** basis dimensions with coefficients shared across all observations.
- PR #55 owns the public-source inward-spiral / axial-stretch feature vector.
- PR #48 owns the static bounded serializable Eq. (4.5) `Phi/F` spatial profile basis.

The remaining Agent-7 question is the user's explicit `time-dependent basis coefficients` option: before doubling or broadly time-expanding the Eq. (4.5) profile family, can exactly **one affine-in-time coefficient lift on one existing spatial mode** add material joint capacity across several reference times?

### Increment
Adds `constrained_temporal_basis_lift.py` with `diagnose_affine_temporal_lift(...)`.

For caller-supplied local residual and visualization response matrices at at least three strictly increasing reference times it compares:
1. a static model with one shared bounded coefficient-update vector at every time; and
2. the same model plus exactly one scalar slope on a declared existing basis direction,
   `delta(t) = delta + tau(t) * slope * e_k`, with `tau` centered/scaled to `[-1,1]` over the supplied time interval.

Both variants use one shared bounded `scipy.optimize.lsq_linear` solve against residual and visualization targets together. The report includes residual/visual/joint remaining ratios, recovered fractions, numerical rank/condition, active bounds, and the gains attributable only to the one temporal lift. Explicit row scales prevent unit-magnitude accidents.

The diagnostic fail-closes on malformed/nonfinite response arrays, non-monotone or fewer-than-three times, mismatched parameter columns, invalid scales/bounds/index, or nonpositive slope budget. It carries `claim_scope=local_expression_capacity_only`, `pde_validated=false`, `visual_correspondence_verified=false`, `paper_exact=false`, and `openai_field_identified=false`.

### Actual local verification
Executed against the final module/test text before upload:
[...]

PR #39, opened 2026-09-16
CR003: add bounded visualization-feature sensitivity diagnostic

Liang-techh:
Constrained Agent 7 minimal expression-capacity increment, branched directly from latest `main@c6700851c0330c0c92938d25204d07fbabf588f4` after rereading README/AGENTS/PROJECT_GOAL/CURRENT_CHECKPOINT/AGENT_TASKS and checking the current open PRs, recent commits, and CI scope.

### task_id
`CR003-VISUAL-FEATURE-SENSITIVITY-006`

### Why this is not duplicate work
- PR #20 provides a generic residual-Jacobian RRQR capacity screen; it does not measure whether basis directions can move visualization morphology.
- PR #28 and PR #35 define candidate-side morphology fingerprints; they do not diagnose parameter/basis sensitivity of those features.
- PR #32 owns Eq. (4.5) incompressible `Phi/F` profile coupling; this PR does not implement q, profiles, or production velocity.
- Agent 1 still owns the callable Eq. (4.5) candidate itself.

### Increment
Adds `constrained_visual_feature_sensitivity.py`, a candidate-agnostic local linearization tool for a declared visualization fingerprint `feature_fn(theta)`.

It:
- finite-differences bounded candidate parameters without ever stepping outside explicit bounds;
- prefers central differences and uses one-sided stencils only at bounds;
- requires explicit positive feature scales so morphology metrics with different physical units are not ranked by raw magnitude;
- forms a dimensionless local feature-response matrix, reports singular values, numerical rank and finite-rank condition number;
- reports each parameter direction's normalized visual-response norm, each feature's local reachability norm, and locally dead parameter labels;
- fail-closes on malformed/nonfinite features, bounds, scales, labels, steps, or parameter states;
- evaluates no PDE residual and introduces no optimizer, forcing, pressure, or acceptance threshold.

The checked synthetic calibration deliberately includes two duplicate visual directions plus one locally dead direction. The diagnostic reports numerical rank 1, condition number 1.0, and correctly identifies the dead parameter. This is calibration only, not evidence about the production candidate.

### Actual verification
Local isolated command actually run before upload:
`PYTHONPATH=/mnt/data/agent7_visual_sensitivity/src python -m pytest -q -W error /mnt/data/agent7_visual_sensitivity/tests/test_constrained_visual_feature_sensitivity.py`
Result: **4 passed in 0.06s**. Python exited 0; an unrelated spreadsheet-runtime warmup traceback was emitted after startup by the automation environment.

[...]

PR #80, opened 2026-09-17
CR003: screen targeted Eq45 morphology correction modes

Liang-techh:
## Task
`CR003-EQ45-TARGETED-MODE-SCREEN-011` — one candidate-specific Constrained Agent 7 basis-growth increment.

## Base / why stacked
This PR is intentionally stacked on open PR #73 (`39bdea136ef09cec5e1eed2391dedae8e1ef40e0`) because #73 is the first frozen, serializable `Eq45VelocityCandidate`. The active integration branch does not yet contain that candidate. This keeps the diff limited to the new expression-capacity experiment rather than duplicating candidate composition.

## Why this is not duplicate work
- #39 owns the generic visualization-feature finite-difference tool.
- #47 owns a generic bounded dual-objective local basis screen.
- #63 owns a generic one-parameter affine time lift.
- #72 audits low-rank separability of the selected `PaperCoreSeries` inner seed.
- #57 owns the physical-space support taper.

This increment is the first screen run directly on the frozen #73 Eq. (4.5) public `[u,v,w]` candidate and asks which *minimal new symmetry-preserving spatial modes* actually provide independent morphology control.

## Increment
Adds `constrained_eq45_targeted_mode_screen.py` plus a checked seed report. The current `(radial_degree=1, eta_degree=2)` Phi/F basis is embedded exactly in `(2,4)`, with every existing coefficient unchanged and every untested new coefficient zero. Only four added directions are perturbed:
- `Phi (2,0)` radial/poloidal;
- `Phi (0,4)` even-in-eta axial/poloidal;
- `F (2,0)` radial/swirl;
- `F (0,4)` even-in-eta axial/swirl.

At `t=0.25/0.5/0.75`, all diagnostics are obtained through `Eq45VelocityCandidate.at_points(...)` on symmetric cylindrical ring probes. The dimensionless feature vector is: radial-thickness ratio, axial-reach ratio, shoulder swirl-energy fraction, and shoulder axial-poloidal fraction. These are autonomous visualization diagnostics, not source facts or acceptance thresholds.

Central coefficient differences use `1e-3` and an independent refinement `5e-4`. The four response columns have singular values approximately
`0.0655995, 0.0413242, 0.0320917, 0.0279838`, so rank is 4 at `rtol=1e-8` and condition number is about `2.344`. The two finite-difference scales differ by only `6.47e-8` in relative Frobenius norm.

Candidate-specific routing result:
- `Phi eta^4` increases axial reach at all three reference times and increases axial-poloidal content, while its midplane radial-thickness response is zero at this local probe by symmetry.
- `F eta^4` independently increases shoulder swirl fraction at all three times and leaves the poloidal split unchanged.
- therefore, if the full visualization proves axially too short, the smallest targeted spatial growth is the Phi eta^4 direction; if that makes the helix too poloidal, pair it with the F eta^4 direction rather than blanket-growing every radial/angular mode.

This is a local reachability/sensitivity result, not a fitted improvement or a claim that any hidden OpenAI target value is known.

## Direct contribution to final `[u,v,w]`
This PR does not change the frozen velocity coefficients. It reduces the next representation-growth search from generic degree increases to two interpretable, well-conditioned axial correction channels that can be activated only if full vorticity/streamline morphology demonstrates an axial-shortness/swirl-balance defect.

## Truth boundary
`velocity_changed=false`, `visualization_ready=false`, `visual_correspondence_verified=false`, `pde_validated=false`, `paper_exact=false`, `openai_field_identified=false`, `blowup_proved=false`.

## Verification
The new regression covers exact zero-extension equivalence, four-direction rank/conditioning and derivative-refinement stability, checked artifact replay, channel-specific axial responses, and fail-closed invalid growth controls. Exact-head GitHub Actions will be recorded after completion; no CI success is claimed here prematurely.

Current head: `1578e97ca5ac47125469d2c885fca63e57bbfc27`
acceptance: pending
merge_status: open


Liang-techh:
Exact-head verification complete for `1578e97ca5ac47125469d2c885fca63e57bbfc27`.

- GitHub Actions run `35169711407`: **completed / success**.
- Fresh Ubuntu 24.04 / CPython 3.11.16 constrained job ran `python -m pytest -q -W error tests/test_constrained_*.py` -> **83 passed in 8.33s**.
- `slice-velocity`, `slice-forcing`, `slice-coordinates`, and `import-and-cli-help` all completed successfully; legacy full matrix was skipped by existing workflow policy.
- PR now reports `mergeable=true` against stacked base #73.

Numerical result remains the checked artifact: singular values `[0.06559949, 0.04132420, 0.03209166, 0.02798381]`, rank 4, condition `2.34419`, max response-column cosine `0.60536`, and `1e-3 -> 5e-4` derivative-refinement relative change `6.47e-8`.

No truth-boundary promotion: this is local morphology reachability only; `velocity_changed=false`, `visualization_ready=false`, `visual_correspondence_verified=false`, `pde_validated=false`, `paper_exact=false`, `openai_field_identified=false`.

PR #128, opened 2026-09-17
CR004: audit restricted-force overlap with Eq45 support taper

Liang-techh:
Task: `CR004-EQ45-SUPPORT-FORCE-OVERLAP-018`

Base: active integration `codex/cr001-constraints@34f8ec692eeeb5b22e89a82b9e4dd6dce9f8721a` after the Eq45 physical-support composition was integrated.

This is one minimal pressure/forcing-governance increment. It does not refit pressure or forcing and does not change `[u,v,w]`. It checks whether the preregistered two-parameter `RestrictedForce(a,c)` remains geometrically compatible with the newly support-connected Eq45 child.

What is checked:
- exact zero of both force basis channels at/outside the same `r=2, |z|=2` cylinder used by `AxisymmetricPhysicalTaper`;
- independent cylindrical Gauss-Legendre L2 quadrature at orders 24/32/48, evaluating only through public `RestrictedForce.__call__`;
- fraction of each force basis energy inside the unchanged taper plateau (`r<=1.6, |z|<=1.6`) versus the support-transform collar.

Checked receipt: `artifacts/constrained/eq45_supported_force_overlap_audit.json`.

Key result at order 48:
- `a`/poloidal channel: 24.584007% of force L2 energy is in the taper collar;
- `c`/swirl channel: 50.485495% is in the taper collar.
Order 32->48 collar-fraction changes are only `2.77e-6` and `2.78e-7`, respectively.

Interpretation: the outer support is compatible, but previously fitted force coefficients on the untapered parent are **not transferable without supported-child re-fit/revalidation**, because a material fraction of the force acts exactly where the support transform changes velocity. This is especially important for c-dominant fits. The next Agent-2 optimization should re-project only the same preregistered `(a,c)` family after an independent supported-child residual operator is available; no residual-defined force direction is justified.

Truth boundary: `velocity_changed=false`, `force_family_changed=false`, `force_refit_performed=false`, `pde_validated=false`, visualization readiness/correspondence not promoted, `paper_exact=false`, `openai_field_identified=false`.


Liang-techh:
Exact-head verification for `15f0f981947005cdd74ead4bf90fa4851accfcf0` is green. Pull-request Actions run `35194877647` completed `success`; push run `35194853810` also completed `success`. The constrained job ran the repository workflow command `python -m pytest -q -W error tests/test_constrained_*.py` and completed successfully; `slice-velocity`, `slice-forcing`, `slice-coordinates`, and `import-and-cli-help` also completed successfully. Legacy full matrix remained skipped by the existing workflow policy. The connector does not expose pytest stdout here, so no pass-count is claimed.

PR #252, opened 2026-09-17
Kokuno A2: remeasure retained curl on reference continuation

Liang-techh:
## Task
`K2-OSC-005` — Kokuno Agent 2, one minimal transfer diagnostic after the routed core checkpoint.

## Fresh audit / non-duplication
Before choosing scope I reread current `README.md`, `AGENTS.md`, `docs/PROJECT_GOAL.md`, live `docs/CURRENT_CHECKPOINT.md`, `docs/AGENT_TASKS.md`, and `docs/SCHEDULED_AGENT_TASKS.md`; audited the current `codex/cr001-constraints` head, open PRs, central Issue #15 claims, recent commits and CI.

Current Kokuno ownership remains separated: Agent 1 owns leading/reference/outer reconstruction (#229/#238/#251), Agent 3 owns mean/radial correction (#241), Agent 4 owns independent validation (#242), and Agent 5 owns integration/checkpoint packaging (#243). Agent-2 #239 already rejected a training-selected scalar phase shift and retained amplitude `.125`, phase `0` on the finite leading core. No current PR remeasures that retained oscillatory correction on Agent-1's new source **reference-continuation** stage.

This branch starts from Agent-5 #243 exact head `1c935253b339dfeb273ca795ba6545d6920cff28` and cross-consumes Agent-1 #251 exact head `035ef85a2d48d00e907afcbace20932542ee0b6a` by exact Git blobs/merge ancestry rather than reimplementing its formulas.

## One minimal increment
Adds `kokuno_reference_oscillatory_screen.py` plus governed regression/workflow. It freezes the already-routed Agent-2 complete-curl surrogate at:
- amplitude `0.125`;
- phase `0.0`;
- unchanged wave vector, polarization, omega, center and compact C4 box support.

No oscillatory parameter is selected in this PR. The screen compares:
1. Agent-1 `KokunoReferenceContinuationCandidate` leading stage alone;
2. the same public leading velocity plus Agent-2's exact-curl correction;

using Agent-4's existing independent black-box FD4 `evaluate_fd4`, Agent-1 pressure unchanged, `nu=.01`, and `f=0` only as a raw local diagnostic.

Held-out evaluation is deterministic and predeclared: seed `9172981`, inner + post-core-X strata, times `.375/.5/.625`, steps `.02/.01/.005`. The post-core stratum is explicitly required to lie beyond the old finite core's `Lambda*X<=4.1` guard, so this checks transfer to the new reference-continuation domain rather than replaying the previous core-only evidence.

[...]

PR #251, opened 2026-09-17
Kokuno: add executable reference continuation candidate

Liang-techh:
## Task
`KOKUNO-LP-REFERENCE-CONTINUATION-005` — Kokuno Agent 1.

## Fresh audit / non-duplication
Before choosing scope I reread current main and live scientific `README.md`, `AGENTS.md`, `docs/PROJECT_GOAL.md`, `docs/CURRENT_CHECKPOINT.md`, `docs/AGENT_TASKS.md`; audited open Kokuno PRs, current integration head/recent commits and CI. Agent 2 owns oscillatory corrections, Agent 3 mean/radial correction, Agent 4 independent validation, and Agent 5 integration/checkpoint packaging. Agent 5 #243 still records Agent 1's source outer profile / compensated core-to-heat assembly as the leading mathematical blocker. No open Kokuno PR owns the source reference-continuation stage implemented here.

This PR stacks directly on Agent-1 #238 exact head `5c7ffdb9851e03d86394ed53c13043278473bb6b`, so it reuses the callable finite core and heat-exterior components without rewriting them.

## Pinned public source / formula
Formula provenance remains pinned to `KokunoYumeto/yang-mills-interacting-workbench@143f6773feb424ad9ed3a8d116653200f20346b7`, `navier-stokes/navier_stokes_workbench.tex`, corrected 2026-09-09 release / Zenodo `22678406`.

The source continuation section fixes `X0=4/Lambda`, `y=log(X/X0)` and the flat step

`sigma(s)=exp(-1/s^2)/(exp(-1/s^2)+exp(-1/(1-s)^2))`

on `0<s<1`, extended by 0/1 outside. For `0<t1<=bar_t` with `4 exp(2 bar_t)<4.1`, it keeps the natural/core profile through `y=t1`, then on `(t1,2t1)` prescribes

`d_y log(phi_r)=(1-sigma((y-t1)/t1))*D_X log(phi_nat)`

`d_y U_r=(1-sigma((y-t1)/t1))*D_X U_nat`, `D_X=X d_X`,

and freezes `phi_r,U_r` for `y>=2t1`. With the source map `F=phi/C`, this is directly executable from the existing `PaperCoreSeries` derivatives.

## Minimal executable increment
Adds `KokunoReferenceContinuationCandidate` and governed regressions. It:
- reuses #229's finite nonlinear natural/core `F,U,Pi` and Kokuno-native q/X/eta coordinates;
[...]

PR #241, opened 2026-09-17
Kokuno A3: remeasure mean correction on actual leading core

Liang-techh:
## Task
`KOKUNO-A3-CORE-MEAN-CORRECTION-CYCLE-004` — Kokuno Agent 3.

One minimal increment after #232/#233: move Agent-3's mean/radial correction machinery off the temporary capped-bipolar bridge and remeasure it on the actual Agent-1 source-native leading-core velocity already integrated with Agent-2's complete-curl correction in #233.

## Fresh audit / non-duplication
Before implementation I reread the current README, AGENTS, PROJECT_GOAL, CURRENT_CHECKPOINT and AGENT_TASKS; audited open PRs, recent commits and current Kokuno CI/results. Agent 1 owns leading/core/exterior reconstruction (#229/#238), Agent 2 owns oscillatory curl/phase screening (#239), Agent 4 owns independent validation, and Agent 5 owns integration packaging (#233). This PR does not duplicate those lanes.

It stacks directly on Agent-5 #233 exact head `4b77580b2bedbbc9b047471cb77b2229549192f2`, because #233 explicitly records the missing seam: Agent-3's #232 correction was selected on a different capped-bipolar bridge and must be remeasured/reselected against Agent-1's actual leading core before composition.

## One minimal increment
Adds `kokuno_core_mean_correction_cycle.py` plus focused guards and a dedicated report workflow.

The cycle fixes Agent-2's currently integrated oscillatory amplitude at `0.125` and phase at `0.0` (Agent-2 #239 rejected its training-selected phase offset on heldout). It does not optimize any Agent-2 parameter.

It then:
1. directly evaluates the real phase-mean nonlinear defect of `KokunoLeadingCoreSeriesCandidate + KokunoCompleteCurlCorrection` through the existing independent CR006 FD operator;
2. reconstructs the theta/e=2 compact radial stress with Kokuno's weighted moment-complement inverse on a predeclared core-safe annulus `[0.08,0.36]`;
3. reuses #232's explicit autonomous stress-shaped mean-swirl lift, but with core-local axial support `|z|<0.20`;
4. screens exactly the unchanged #232 nonzero amplitude grid `[-.04,-.02,-.01,-.005,.005,.01,.02,.04]` on held-in seed `9172921` only;
5. freezes the selected amplitude and evaluates it on disjoint held-out seed `9172927` at `t=.375/.5/.625`;
[...]

PR #253, opened 2026-09-17
Kokuno A3: measure signed covariance inverse on real radial stress

Liang-techh:
## Task
`KOKUNO-A3-SIGNED-COVARIANCE-INVERSE-005` — Kokuno Agent 3.

One minimal increment after #241/#243: replace the rejected direct `radial stress -> mean swirl velocity` routing with an executable **signed differential covariance** realizability diagnostic driven by the actual leading-core mean defect and the already routed Agent-2 complete-curl column.

Before implementation I reread the live README, AGENTS, PROJECT_GOAL, CURRENT_CHECKPOINT and AGENT_TASKS; audited open PRs, recent commits and CI. Agent 1 #251 owns the reference continuation / remaining outer-profile reconstruction. Agent 2 #252 owns transfer of the retained oscillatory correction onto that continuation. Agent 5 #243 is the current routed checkpoint and explicitly records Agent 3's remaining blocker: replace/complete the stress-to-velocity realization. This PR does not implement a new curl kernel, select Agent-2 phase/amplitude, or duplicate Agent-1 outer work.

It stacks on Agent-5 #243 exact head `1c935253b339dfeb273ca795ba6545d6920cff28`, so the actual leading core, routed Agent-2 amplitude `0.125` / phase `0`, prior Agent-3 rejection, and Agent-4 evidence are in ancestry.

## Pinned source structure
The corrected public reader is pinned to `KokunoYumeto/yang-mills-interacting-workbench@143f6773feb424ad9ed3a8d116653200f20346b7`, `navier-stokes/navier_stokes_workbench.tex`, corrected 2026-09-09 / DOI `10.5281/zenodo.22678406`.

Its signed differential inverse keeps the original positive primary amplitudes fixed and uses

`d_Sigma = H^{-1}(Sigma/epsilon)`

`delta a_sigma = (d_Sigma)_sigma / (2 a_sigma)`

`L Sigma = sqrt(epsilon) sum_sigma delta a_sigma b_sigma`

with exact linear covariance identity `B(W0,L Sigma)=Sigma`. The source explicitly does **not** take a square root of a perturbed covariance; the nonlinear self-covariance is retained separately.

The current repository does not expose the source two-column `H`, `a_sigma`, or pulse columns. This PR therefore does not pretend to implement the source map exactly. It measures the physical rank-one analogue directly from Agent 2's public velocity:

`d/da <w_r w_theta>|a0 = 2 a0 <w_unit,r w_unit,theta>`.

Adds `kokuno_signed_covariance_inverse.py`, focused tests and a dedicated Actions report. The report:
1. evaluates the **real** phase-mean defect of `KokunoLeadingCoreSeriesCandidate + KokunoCompleteCurlCorrection` through the existing independent FD operator;
2. reconstructs the compact theta/e=2 radial stress on the predeclared core annulus `[0.08,0.36]` using the existing weighted moment-complement inverse;
3. measures the routed Agent-2 column's radial-theta covariance response by public `at_points` calls over 8 physical angles and 8 phase shifts;
4. solves the scalar signed differential-amplitude analogue only where the measured covariance response is safely nonzero;
5. reports unresolved active stress nodes, response conditioning, required signed amplitude, exact compact edge values, and an independent finite-difference calibration of the quadratic-in-amplitude covariance law;
6. fails closed as a routing result if the one-column response loses rank or exceeds the existing Agent-2 absolute amplitude bound.

No amplitude range is widened after seeing the result. No pressure/force is fitted. No hand-entered defect is accepted.

## Scientific boundary
This is a **covariance-realizability increment**, not yet a public velocity correction. Even if the rank-one map is well conditioned, the next step must materialize that signed amplitude at vector-potential level through the existing complete curl and then re-run disjoint held-in/held-out residual checks. If the one-column map is rank-deficient, the correct next step is a second independent fixed positive column / source-H analogue, not scalar amplitude inflation.

`real_candidate_defect_consumed=true`, `radial_stress_reconstructed_from_real_defect=true`, `source_signed_differential_formula_recorded=true`, but `source_exact_signed_map_implemented=false`, `public_velocity_correction_materialized=false`, `residual_reduction_claimed=false`, `formal_full_domain_normalized_pde_gate_assessed=false`, `normalized_ns_residual_le_1e-3_claimed=false`, `pde_validated=false`, `paper_exact=false`, `openai_field_identified=false`, `blowup_proved=false`.

Exact-head numerical results and CI run IDs will be appended only after Actions resolves; no result is fabricated in this opening description.


Liang-techh:
Exact-head verification / scientific result for `17f15998218f885b072e2e483de4aa26f2f80268`:

- Dedicated Actions run `35278036467`: **success** on Ubuntu 24.04.5 / CPython 3.11.16.
- Focused command: `python -m pytest -q -W error tests/test_constrained_kokuno_signed_covariance_inverse.py tests/test_constrained_kokuno_radial_stress.py tests/test_constrained_kokuno_mean_defect.py tests/test_constrained_kokuno_complete_curl.py` -> **19 passed in 1.06s**.
- Real report command: `python -m openai_ns_reconstruction.kokuno_signed_covariance_inverse --output artifacts/kokuno_agent3/signed_covariance_inverse_report.json --radial-count 33 --angular-count 8 --phase-count 8` -> success.
- Standard repository tests run `35278036364`: **success**.
- Report artifact ID `10521700003`, ZIP digest `sha256:b634cb07afcf07f3698fd8cda0a67573e89090af618016b093aabdf2e551b853`.

Actual-core / actual-defect numbers (`t=.5`, `z=0`, annulus `[.08,.36]`, routed Agent-2 amplitude `.125`, phase `0`):
- full real phase-mean defect increment RMS: `1.0424882938e-2`;
- raw theta ring-mean RMS: `4.6021014524e-4`; fixed annular gate capture ratio `0.57020318`;
- compact theta stress RMS/max: `2.2592000498e-6 / 4.6554161147e-6`;
- equal-moment radial inverse identity FD RMS/max: `9.9016529123e-13 / 1.9132722276e-12`; outer/off-annulus stress remains exactly zero; without moment subtraction the implied outer tail is `1.8241853571e-5`;
- measured unit `<w_r w_theta>` RMS: `2.3771928156e-3`;
- measured covariance derivative `d/da <w_r w_theta>|a=.125` RMS/max: `5.9429820390e-4 / 7.4546315926e-4`;
- independent amplitude-FD calibration relative error: `2.3375100190e-14`;
- active target nodes `26`, safe response nodes `26`, unresolved active nodes `0`; response condition ratio `2.48147`;
- required signed differential amplitude RMS/max: `3.2610377627e-3 / 6.6356147307e-3`, with exact zero edge values and within the existing Agent-2 absolute amplitude bound;
- sampled linear covariance reconstruction relative RMS: `1.6570579124e-17`.

Routing result: **the current routed Agent-2 column is not rank-deficient for this measured theta stress on the predeclared core annulus**, so the signed-covariance route is eligible for the *next* increment: materialize this radial signed amplitude at vector-potential level using the existing complete curl, then recompute disjoint held-in/held-out real residual before accepting any correction. This PR itself does not materialize `delta u`, does not claim residual contraction, and does not assess the formal full-domain normalized `<=1e-3` gate.

Source boundary remains explicit: Kokuno's exact two-column map uses `d_Sigma=H^{-1}(Sigma/epsilon)`, `delta a_sigma=(d_Sigma)_sigma/(2a_sigma)` and `B(W0,L Sigma)=Sigma`; the repository currently measures a one-column physical analogue because source `H/a_sigma/b_sigma` are not yet exposed. No free force, fitted pressure, surrogate defect, or threshold change is used.

Liang-techh:
Agent-4 dependency feedback from independent #254: the current executable reference-continuation leading baseline is not yet suitable for promotion testing of a realized covariance correction. Separate FD2 black-box validation finds post-core `t=.625` raw L2 `110.4742`, max `141.7033`, radial RMS `102.2291`, and divergence max `0.35520`, stable from h=.002→.001. Your signed-covariance/rank-one realizability diagnostic can continue as a structural lane, but once a public correction velocity exists, full before/after residual should be interpreted against this leading/global-matching obstruction; a correction that only moves the mean/stress channel cannot by itself certify PDE progress while leading divergence remains O(1e-1). Formal 1e-3 gate remains unassessed.

PR #280, opened 2026-09-18
Kokuno: compute source heat-replacement discrepancy

Liang-techh:
## Task
`KOKUNO-LP-ACTUAL-HEAT-DISCREPANCY-008` — Kokuno Agent 1.

## Fresh audit / non-duplication
Before choosing scope I reread latest main and live scientific `README.md`, `AGENTS.md`, `docs/PROJECT_GOAL.md`, `docs/CURRENT_CHECKPOINT.md`, `docs/AGENT_TASKS.md`, checked current live head `3c22198881d5f46e04a1f6551f954eb73027cd07`, open Kokuno PRs, recent integration commits and CI. Agent 4 #274 independently preflights #271; Agent 5 #276 explicitly routes the next leading-profile blocker to Agent 1: supply the source heat-replacement discrepancy/scales and feed the bounded repair. No other Kokuno PR owns this calculation.

This PR stacks directly on Agent-1 #271 exact head `d2b20eeb52e1914e2aaa99be9ef6437d819a3e53`.

## Pinned public source / formula map
Provenance remains pinned to `KokunoYumeto/yang-mills-interacting-workbench@143f6773feb424ad9ed3a8d116653200f20346b7`, `navier-stokes/navier_stokes_workbench.tex`, corrected 2026-09-09 / Zenodo `22678406`.

The source heat-replacement section sets `y=log(X/X_tail)`, `X_K=e^.2 X_tail`, `X_b=e^3 X_tail`, chooses `chi_K=0` for `y<=.2` and `chi_K=1` for `y>=.5`, and replaces `E_cl` by
`E_cl[1+chi_K(Hcal(2(1-eta^2)/X)-1)]`.
On the terminal tail `E_cl=c_inf X^(-A) f_o`, `A=1/2+h`, with `f_o=1-rho_o psi_o`, `psi_o=1-sigma((y-1)/2)`, `rho_o=c_o h>0`, `0<=f_o'/f_o<h/4`.
The induced three discrepancies are exactly from `delta(E^2)/(2X)`, `-delta(E^2)/2`, and `sqrt(2X) delta E`.

The source fixes only the endpoint/support contract for `chi_K`; this implementation uses the explicit autonomous flat realization `sigma((y-.2)/.3)`. Numerical `X_tail,c_inf,rho_o` remain explicit inputs and are **not** claimed to be recovered source hidden choices.

## Minimal executable increment
Adds `KokunoHeatReplacementDiscrepancy` with:
[...]

PR #275, opened 2026-09-17
Kokuno A3: reject phase copies as second covariance column

Liang-techh:
## Task
`KOKUNO-A3-PHASE-ORBIT-COVARIANCE-RANK-007` — Kokuno Agent 3, one minimal source-structure/routing increment after #263 and Agent-5 independent routing #268.

## Fresh audit / non-duplication
Before implementation I reread the live `README.md`, `AGENTS.md`, `docs/PROJECT_GOAL.md`, `docs/CURRENT_CHECKPOINT.md`, `docs/AGENT_TASKS.md`; audited the live `codex/cr001-constraints` head (`a918e0f6b0f6fd05d1cd65252939c18af3cde6bb`), current open PRs, recent commits and CI. Agent 1 #271 still owns leading/heat repair and does not yet provide a globally matched leading field. Agent 2 #272 now bridges the leading profile into the source phase chart but does not yet materialize the source two-label complete-curl pair. Agent 4 #264 and Agent 5 #268 independently reject routing/retuning Agent-3 #263's one-column signed correction. This PR does not implement a new oscillatory kernel, polarization, wave vector, amplitude, phase selection, leading profile or independent PDE validator.

It stacks directly on Agent-5 #268 exact head `6e78e34e0eb1b30cc19ad09ef47ea8e6318ac7e1`, which already binds the negative independent result for #263.

## Pinned source requirement
Provenance remains `KokunoYumeto/yang-mills-interacting-workbench@143f6773feb424ad9ed3a8d116653200f20346b7`, `navier-stokes/navier_stokes_workbench.tex`, corrected 2026-09-09 / Zenodo `22678406`.

The corrected reader uses two distinct fixed positive pulse columns

`H_sigma = h_sigma (-A_c N - sigma u_* K + e_sigma)`

and, before the small column errors,

`H_ref=[[-A_c h_+, -A_c h_-],[-u_* h_+, +u_* h_-]]`,

with exact reference determinant `-2 A_c u_* h_+ h_- != 0`. Its signed differential inverse is

`d_Sigma=H^{-1}(Sigma/epsilon)`, `delta a_sigma=(d_Sigma)_sigma/(2a_sigma)`,

with `B(W0,L Sigma)=Sigma` and the nonlinear self-covariance `C(L Sigma)` retained separately.

## One minimal executable increment
Adds `kokuno_phase_orbit_covariance_rank.py` plus focused regressions and a dedicated report workflow. It:
[...]

PR #272, opened 2026-09-17
Kokuno A2: bridge leading profiles into source phase chart

Liang-techh:
## Task
`K2-OSC-008` — Kokuno Agent 2, one minimal source-structure increment after #262.

## Fresh audit / non-duplication
Before implementation I reread the live `README.md`, `AGENTS.md`, `docs/PROJECT_GOAL.md`, `docs/CURRENT_CHECKPOINT.md`, and `docs/AGENT_TASKS.md`; audited the live `codex/cr001-constraints` head, current open PRs, Issue #15 claims, recent commits and available CI. Agent 1 owns leading/reference/heat repair, Agent 3 owns mean/radial correction, Agent 4 owns independent validation, and Agent 5 owns integration/frontier receipts. Agent 5 still records the Agent-2 source-phase seam as blocked on the base `V/G` mapping. No active PR owns the leading-chart scaling bridge implemented here.

This PR stacks directly on Agent-2 #262 exact head `71078654aadadcab1549189c06c2737e4595544f` and does not alter #255's complete-curl velocity.

## Pinned source formulas and derivation
Provenance remains `KokunoYumeto/yang-mills-interacting-workbench@143f6773feb424ad9ed3a8d116653200f20346b7`, `navier-stokes/navier_stokes_workbench.tex`, corrected 2026-09-09 / Zenodo `22678406`.

The oscillation chapter uses the dyadic chart

`R=r/Q^(1/2), Z=z/Q^D, T=(1-t)/Q, epsilon=Q^h`, with physical velocity `Q^(-A) u_*`, and writes the chart tangential fields as `(V,G)` with `F_chart=V/R`.

Agent-1's source-leading profile is represented by

`u_theta=q^(-A)*sqrt(2X)*F_profile`, `u_z=q^(-A)*U`, and `s=q/Q`.

Therefore the source scaling itself gives the **leading-only** bridge

`F_chart = s^(-(A+1/2))*F_profile`, `V=R*F_chart`, `G=s^(-A)*U`.

The module differentiates this map analytically in chart `R,Z`, reusing Agent-1's native coordinate Jacobian for `s,eta,X`, and feeds the resulting `V,V_R,V_Z,G,G_R,G_Z` into #262's executable source phase/covector contract.

## One minimal executable increment
Adds `KokunoAgent1LeadingChartPhaseBridge`:
- bounded dyadic replay label `ell` and enforced `epsilon=Q^h`;
- executable leading chart `V,G` plus analytic `R/Z` derivatives;
[...]

PR #263, opened 2026-09-17
Kokuno A3: materialize signed amplitude through complete curl

Liang-techh:
## Task
`KOKUNO-A3-SIGNED-AMPLITUDE-CURL-CYCLE-006` — Kokuno Agent 3.

One minimal increment after #253: turn the measured, well-conditioned signed radial amplitude profile into an actual public velocity correction at **vector-potential level**, then recompute the real phase-mean NS defect on fresh held-in and disjoint held-out points. This is the first test of whether #253's covariance-level realizability survives the full complete-curl lift.

## Fresh audit / non-duplication
Before implementation I reread the latest live `README.md`, `AGENTS.md`, `docs/PROJECT_GOAL.md`, `docs/CURRENT_CHECKPOINT.md`, and `docs/AGENT_TASKS.md`; audited current open PRs, recent live commits and CI. Agent 1 #261 owns five-moment/leading-profile repair. Agent 2 #255 owns native-X oscillatory localization/phase-frame work. Agent 5 #256 explicitly records Agent 3's remaining blocker as materializing #253's signed differential amplitude and rerunning held-in/held-out residual checks. This PR does not implement a new oscillatory kernel, select Agent-2 amplitude/phase, or duplicate Agent-1 outer/heat work.

It stacks directly on Agent-3 #253 exact head `17f15998218f885b072e2e483de4aa26f2f80268`. Final exact head is `04f1335dd73eecffb95d18c523f1afb81ee89195`.

## Source structure and velocity lift
Pinned source remains KokunoYumeto corrected 208-page reader, `2026.09.09-consolidated`, DOI `10.5281/zenodo.22678406`, source commit `143f6773feb424ad9ed3a8d116653200f20346b7`.

#253 records the source signed differential map

`d_Sigma = H^{-1}(Sigma/epsilon)`

`delta a_sigma = (d_Sigma)_sigma/(2 a_sigma)`

and measured a rank-one physical analogue against Agent 2's already-routed complete-curl column. The repository still lacks the source two-column `H`/pulse hierarchy, so this PR does not claim source exactness.

For the public velocity materialization, this PR reuses Agent 2's exact public vector potential and curl. If `A_unit` is Agent 2's amplitude-one potential and `w_unit=curl A_unit`, then for the signed radial amplitude `alpha(r)`:

`A_delta = alpha(r) A_unit`

`delta_u = curl(A_delta) = alpha w_unit + grad(alpha) x A_unit`.

[...]

PR #282, opened 2026-09-18
Kokuno A3: expose real missing covariance column target

Liang-techh:
## Task
`KOKUNO-A3-MISSING-COVARIANCE-COLUMN-TARGET-008` — Kokuno Agent 3, one minimal routing/typed-contract increment after #275.

Before implementation I reread the live `README.md`, `AGENTS.md`, `docs/PROJECT_GOAL.md`, `docs/CURRENT_CHECKPOINT.md`, and `docs/AGENT_TASKS.md`; audited live `codex/cr001-constraints` head `3c22198881d5f46e04a1f6551f954eb73027cd07`, open PRs, recent commits, and exact-head CI for #275. Agent 1 has advanced through #280 heat-replacement discrepancy machinery but still does not expose a globally matched leading field. Agent 2's latest #272 makes the source phase/chart bridge executable but still does not materialize the corrected reader's genuinely independent two-label complete-curl pair. Agent 5 #268 explicitly rejects rerouting the one-column #263 signed correction while the rank/direction and leading blockers remain.

This PR stacks directly on Agent-3 #275 exact head `22845a379f38199777114ed93079a20499064006`. It does not construct a new oscillatory pulse, polarization, wave vector, leading profile, pressure/force fit, or PDE validator, and does not rerun the rejected finite correction cycle.

## Why this increment
#275 established globally that phase `0` and `pi/2` copies of the existing Agent-2 column give the same phase-mean quadratic covariance response: rank 1 with about 65.67% of the real two-component compact stress target left outside a *single global scalar* span. The next useful Agent-3 object is therefore not another phase copy or larger amplitude, but an executable **real missing-column response target** that a future genuinely independent Agent-2 public velocity column can be tested against.

Adds `kokuno_missing_covariance_column_target.py` and focused tests/workflow. It:
1. consumes the actual Agent-1 leading core + routed Agent-2 public complete-curl through the existing real phase-mean NS-defect contract;
2. reconstructs both compact theta/e=2 and axial/e=1 radial stresses on the same frozen annulus/time/slice used by #275;
3. measures the current two-component public covariance derivative `2 a (<w_r w_theta>, <w_r w_z>)`;
4. at every radial node, removes the best scalar projection of the real stress target onto that current response;
5. records the orthogonal remainder as the **minimum missing two-component covariance-response target**;
6. exposes `evaluate_second_column_candidate(...)`, which can consume a future measured second public covariance column and report pointwise rank, conditioning, and residual coverage before any finite correction cycle is allowed;
7. calibrates the contract against a duplicate current column (must stay rank deficient) and against the abstract missing remainder itself (algebraic handoff only, never candidate data).

The target uses the existing preregistered `1e-3` relative target/response floors from the signed-covariance lane. It is a measured response requirement, **not** a velocity, source pulse, or recovered hidden parameter.

## Exact-head result
Final head: `56501232583f7e08f02c223e9b9e5b9d6ca94c27`.

The real phase-mean defect receipt reproduces the #275 scale: full mean-defect increment RMS `9.586153581e-3`, ring-theta raw RMS `4.030237855e-4`, ring-z raw RMS `2.174868508e-3`. On the compact theta/axial stress target, 27 nodes are active and all 27 have a safe nonzero current public covariance response.

Allowing the current column its source-motivated **radially varying pointwise coefficient** changes the geometry substantially relative to #275's intentionally harsher single-global-scalar audit: the best current-column projection leaves a measured vector RMS remainder of `2.112128314e-6` against target vector RMS `1.004844700e-5`, i.e. **`0.2101945021` relative RMS**. The remaining direction is predominantly theta: theta RMS `1.912775269e-6`, axial RMS `8.957548695e-7`. The projection is numerically clean: max pointwise cosine between current response and the constructed missing remainder is `1.28137e-15`.

The key routing fact survives the more source-aligned pointwise treatment: **25 of 27 active radial nodes still require a genuinely second direction.** A duplicate of the existing column gives `0/27` rank-two active nodes and leaves the same `0.2101945021` residual. Feeding the abstract orthogonal remainder itself as a calibration-only second response gives `27/27` rank-two nodes and `2.3436e-16` two-column residual, confirming that the contract is algebraically sufficient. That abstract calibration has raw-scale condition median about `1.09e3` and max about `2.51e5`; it is deliberately not interpreted as a realizable physical pulse or as source conditioning.

This sharpens #275 rather than contradicting it: the earlier 65.67% number was the unexplained RMS under one **global** scalar coefficient across the whole radial profile, whereas this PR asks the source-relevant local question and finds that radial amplitude freedom removes much of that mismatch, but not the transverse covariance direction. It also does **not** rescue #263: the previously materialized one-column public correction still failed the real finite-cycle residual test, so this receipt is a handoff target, not permission to retune/rerun the rejected one-column lift.

## CI / reproducibility
Dedicated workflow `35292279365` completed **SUCCESS** on Ubuntu 24.04.5 / CPython 3.11.16. Focused command:

`python -m pytest -q -W error tests/test_constrained_kokuno_phase_orbit_covariance_rank.py tests/test_constrained_kokuno_missing_covariance_column_target.py`

returned **8 passed in 0.62s**. The real report command:

`python -m openai_ns_reconstruction.kokuno_missing_covariance_column_target --output artifacts/kokuno_agent3/missing_covariance_column_target_report.json --radial-count 33 --angular-count 8 --phase-count 8`

completed successfully. Report artifact ID `10527151558`, digest `sha256:56107f975bfcad3157eda8b14fc59fdeba3d2227a26a1930a8637814f6197874`.

Standard repository workflow `35292279284` also completed **SUCCESS**; constrained integration ran `python -m pytest -q -W error tests/test_constrained_*.py` with **223 passed in 81.87s**. Inherited Agent-5 integration checkpoint workflow `35292279335` also completed **SUCCESS**.

## Source/provenance boundary
Pinned provenance remains `KokunoYumeto/yang-mills-interacting-workbench@143f6773feb424ad9ed3a8d116653200f20346b7`, `navier-stokes/navier_stokes_workbench.tex`, corrected 2026-09-09 / Zenodo `22678406`. The source requires two distinct signed covariance columns in `H`; this PR only computes what the current repository's one-column public response cannot span.

## Routing / truth boundary
The next admissible Agent-3 finite-cycle step is gated on Agent 2 materializing a genuinely independent source-motivated public velocity column. Once that exists, Agent 3 can measure its theta/axial covariance response with `evaluate_second_column_candidate(...)`; only a measured rank-two response that improves real target coverage should be lifted into the next public correction and held-in/held-out residual cycle.

`surrogate_defect_used=false`, `real_candidate_defect_consumed=true`, `missing_second_column_target_measured=true`, but `new_oscillatory_column_constructed=false`, `public_velocity_correction_materialized=false`, `finite_correction_cycle_run=false`, `residual_reduction_claimed=false`, `formal_full_domain_normalized_pde_gate_assessed=false`, `pde_validated=false`, `paper_exact=false`, `openai_field_identified=false`, and `blowup_proved=false` remain hard states.


Liang-techh:
## Cross-route baseline alert — ST006 must be visible to all Kokuno agents

Repository `main` already contains the retained ST006 research baseline (merged via #245; publication metadata fixed in #273).

Authoritative independent replay:
- candidate SHA256: `6b4d84b48ab9dbcd2ee1a1858d3e56ef81523f5864369d7e96c6431fccf107a3`
- held-out seed: `9172801`
- 4096 Cartesian points, six fixed times
- FD4 spatial step `h=0.005`
- full three-component momentum residual max: **0.1082289305112118**
- target remains **1e-3**
- divergence_max still fails the registered gate
- `pde_validated=false`

Runnable baseline:
```python
from research_baseline import load_best
field = load_best()
```
and `python scripts/ns_candidate.py validate --seed 9172801 ...`.

### Coordination rule for Kokuno Agents 1–5
Treat ST006 as the **repository-wide retained numerical baseline**, not as Kokuno-derived evidence and not as an accepted NS solution. Every future Kokuno residual report/checkpoint should state whether its metric is directly comparable with the ST006 protocol. If not, label it incomparable rather than treating the current O(10) Kokuno intermediate residual as the best repository result.

For Agent 5 integration specifically: add a baseline-vs-Kokuno reference row in the next replayable checkpoint. For Agent 4: when a global Kokuno candidate exists, run the unchanged formal gate and also report a same-protocol comparison against ST006. Agents 1–3 should use ST006 only as a performance/reference baseline; do not import its coefficients as Kokuno source truth.

This does not change any threshold, forcing contract, or truth state.

PR #281, opened 2026-09-18
Kokuno A2: make source cylindrical complete curl executable

Liang-techh:
## Task
`K2-OSC-009` — Kokuno Agent 2, one minimal source-complete-curl increment after #272.

## Fresh audit / non-duplication
Before implementation I reread the live README/AGENTS/PROJECT_GOAL/CURRENT_CHECKPOINT/AGENT_TASKS, audited `codex/cr001-constraints@3c22198881d5f46e04a1f6551f954eb73027cd07`, current open PRs/recent commits and Issue #15. Agent 1 #280 owns heat/global-leading repair, Agent 3 #275 owns covariance/mean-correction diagnostics, Agent 4 #274 owns independent validation, and Agent 5 #276 explicitly says #272 must remain provenance-ready glue rather than a promoted source-oriented velocity until actual-background V/G positive-order corrections exist.

This PR therefore does **not** compose #272's leading-only V/G into a claimed source wave. It makes only the source stage-9 complete-curl algebra executable for caller-supplied source data.

## Pinned source formulas
Provenance remains `KokunoYumeto/yang-mills-interacting-workbench@143f6773feb424ad9ed3a8d116653200f20346b7`, `navier-stokes/navier_stokes_workbench.tex`, corrected 2026-09-09 / Zenodo `22678406`.

For nonzero harmonic `m`, source stage 9 records

`k=ceil(epsilon^(-1/2))`, `k_m=k*m`, `n_Phi dot t_m=0`,

`C_m = i*(n_Phi x t_m)/(k_m*|n_Phi|^2)`,

`curl_*(C_m exp(i*k_m*Phi)) = (t_m+r_m) exp(i*k_m*Phi)`,

`r_m=(-D_z C_theta, D_z C_r-D_r C_z, (D_r+R^-1)C_theta)`.

The source also retains the complete amplitude's longitudinal remainder; it is not asserted to remain purely transverse.

## One minimal executable increment
Adds `KokunoSourceCompleteCurlContract`:
- exact bounded single-harmonic `C_m` algebra;
- exact cylindrical derivative/frame remainder `r_m`;
- batch complex normalized vector-potential and complete-mode evaluation;
- explicit transverse/covector guards;
- direct `epsilon,m` binding to #262/#272-style phase contracts;
- deterministic JSON/SHA provenance and fail-closed truth metadata.

[...]

PR #245, opened 2026-09-17
Publish retained ST006 to main and organize repository entry points

Liang-techh:
## Completed: retained ST006 published to main; science still NOT accepted

User task PUB-ST006-001 / #205. Merged into main as `a69ebc4dfb427f6a994a2726969e1086b1023d84`, from reviewed head `749999cb093906db32b859eb983cf9e8387da67f`.

### Actual publication
- Exact raw ST006 `artifacts/research/ST006/candidate.json` (1946 stored parameters), three byte-exact original validation/refinement/comparison JSONs and checksum manifest. Original candidate SHA256: `6b4d84b48ab9dbcd2ee1a1858d3e56ef81523f5864369d7e96c6431fccf107a3`.
- Repository-local `research_baseline.load_best()` exposes Cartesian velocity, pressure and original restricted force, with read-only coefficients, broadcasting, validation and chunking. Numerical evaluator arithmetic is unchanged; imports are namespaced.
- `scripts/ns_candidate.py`: status / verify / evaluate / validate. Clean checkout can load and independently validate ST006 without recovering chat attachments or rerunning fitting.
- Exact `experiments/root_st030` implementation from commit `7c63cdb9ca00090acfce42c05ddc512397117e08`, not a wholesale merge of its parent scientific branches.
- Root README, current checkpoint, task routing and project_status corrected from migration-era not_constructed to retained/not_validated. Added document, directory, branch/PR and experiment indexes. Four original entry files preserved as exact Git blobs under `docs/archive/pre_publication/`.
- Dedicated read-only `research-publication` workflow validates immutable artifacts, tests API compatibility and reproduces full independent scientific rejection.

### Actual verification before merge
Local publication regressions: **8 passed in4.63s**; layout and compileall passed.
`python scripts/ns_candidate.py validate --seed 9172801 --out outputs/ST006_local_recheck.json` completed and returned1. All gates matched the original evidence; differences in momentum max/L2 across all spatial rows were0.0 locally.

[...]

PR #283, opened 2026-09-18
Kokuno A4: independently audit heat-replacement discrepancy

Liang-techh:
## Task
`KOKUNO-A4-HEAT-DISCREPANCY-INDEPENDENT-009` — Kokuno Agent 4, one minimal independent-validation increment after Agent 1 #280.

Before implementation I reread latest main and live scientific `README.md`, `AGENTS.md`, `docs/PROJECT_GOAL.md`, `docs/CURRENT_CHECKPOINT.md`, `docs/AGENT_TASKS.md`, and the scheduled routing; audited live `codex/cr001-constraints@3c22198881d5f46e04a1f6551f954eb73027cd07`, open Kokuno PRs, recent integration commits, and latest Agent 1–3 work. Agent 1 #280 owns the actual source heat-replacement discrepancy; Agent 2 #281 owns source complete-curl algebra; Agent 3 #282 owns the missing covariance-column target. This PR stacks only on #280 exact head `51076990abdd36bcb7dc598a7abd85d677dee53d` and does not construct a new leading/oscillatory/correction candidate.

## Independent operator path
The new validator does **not** use #280's Gauss-Legendre/Gauss-Laguerre discrepancy split and does not call `KokunoHeatExteriorProfile` for the heat factor. Instead it uses the source small-z expansion
`(H(z)-1)/z = sum_{m>=1} (-1)^m (h)_m (h+1)_m z^(m-1)/m!`
(valid here because the default tail has `z<=0.002`), adaptive quadrature directly in `y=log(X/X_tail)`, and an independent `s=h(y-3)` tail variable for the long `I_sub` scale.

## New finding
At the public default and `eta=.2`, the independent reference is approximately
`[-2.21476226138e-09, 6.25423009649e-06, -2.63159129494]`.
The public #280 `Delta S` and `Delta I_sub` agree closely, but `Delta C_p` is high by about **0.24325%**. The discrepancy equals the missing half of the first-order `y>=3` `Delta C_p` tail term.

Algebraically, for `H-1=z R(z)` and `z=2d/X`, the first-order `Delta C_p` tail integrand is
`2*c_inf^2*d*X^(-(2A+1))*R(z)`.
#280's tail formula carries only `1*c_inf^2*d*...`, while its second-order term is correctly weighted. The new audit measures `(public-reference)/(-0.5*first_order_tail) ~= 1` across several eta values.

This is a local source-moment bug, not a full PDE verdict. Agent 1 should correct the factor and rerun its own deterministic checkpoint before using the heat discrepancy as the three-bump repair target.

## Truth boundary
No training loss is used; no free forcing is introduced; no velocity/pressure candidate is promoted. `formal_full_domain_pde_gate_assessed=false` and `pde_validated=false` remain unchanged. The fixed `1e-3` full NS gate and `1e-5` divergence gate are not modified.

## Verification
Added focused tests for the independent reference, multi-eta diagnosis, exact missing-half-tail attribution, and series-order stability. Exact-head GitHub Actions results will be recorded only after they resolve.


Liang-techh:
Exact-head CI receipt for `ecf1be298e880afe62b1f568a8f42bd27bc54d4a`: Actions run `35292637019` completed **success**. `constrained-integration`, `slice-velocity`, `slice-forcing`, `slice-coordinates`, and `import-and-cli-help` all completed successfully; the existing legacy full matrix was skipped by policy. PR is `mergeable=true`. Scientific result remains local: the independent audit identifies the missing factor 2 in #280's first-order `Delta C_p` terminal-tail term; no full-domain velocity/pressure composite was available, so the fixed `1e-3` NS gate and `1e-5` divergence gate remain unassessed and `pde_validated=false`.

Liang-techh:
## Cross-route baseline alert — ST006 must be visible to all Kokuno agents

Repository `main` already contains the retained ST006 research baseline (merged via #245; publication metadata fixed in #273).

Authoritative independent replay:
- candidate SHA256: `6b4d84b48ab9dbcd2ee1a1858d3e56ef81523f5864369d7e96c6431fccf107a3`
- held-out seed: `9172801`
- 4096 Cartesian points, six fixed times
- FD4 spatial step `h=0.005`
- full three-component momentum residual max: **0.1082289305112118**
- target remains **1e-3**
- divergence_max still fails the registered gate
- `pde_validated=false`

Runnable baseline:
```python
from research_baseline import load_best
field = load_best()
```
and `python scripts/ns_candidate.py validate --seed 9172801 ...`.

### Coordination rule for Kokuno Agents 1–5
Treat ST006 as the **repository-wide retained numerical baseline**, not as Kokuno-derived evidence and not as an accepted NS solution. Every future Kokuno residual report/checkpoint should state whether its metric is directly comparable with the ST006 protocol. If not, label it incomparable rather than treating the current O(10) Kokuno intermediate residual as the best repository result.

For Agent 5 integration specifically: add a baseline-vs-Kokuno reference row in the next replayable checkpoint. For Agent 4: when a global Kokuno candidate exists, run the unchanged formal gate and also report a same-protocol comparison against ST006. Agents 1–3 should use ST006 only as a performance/reference baseline; do not import its coefficients as Kokuno source truth.

This does not change any threshold, forcing contract, or truth state.

PR #287, opened 2026-09-18
CR003: screen support-adjacent axial-cap poloidal capacity

Liang-techh:
Superseded by clean current-ancestry replay PR #343, which copied exactly this PR's three owned blobs onto live `codex/cr001-constraints@f6f4e15dabf9d4aa63cbf9740b0398dc3c47daf0` after #279 landed. #343 exact replay head CI passed and was squash-merged as live SHA `4858f5e5b17449414d4be0768858e2d2043b5324`.

Scientific scope is unchanged: this remains axial-cap representation/visualization capacity evidence only. Diagnostic ±0.25 is not a production coefficient/bound; canonical velocity/candidate SHA, pressure/force, thresholds and truth states were not promoted.


Liang-techh:
## Repository-wide NS routing update: advance from ST006

The retained ST006 baseline on `main` must now be treated as the repository-level PDE benchmark by **all** active NS agents, not only the Kokuno lane.

Independent replay:
- momentum max: **0.1082289305112118**
- seed `9172801`, 4096 held-out Cartesian points, six fixed times, finest spatial step `0.005`
- original target remains `1e-3`
- divergence gate still fails
- `pde_validated=false`

Routing consequence:
- If your candidate/result is directly comparable to this protocol and is worse than ST006, do **not** present it as repository-level PDE progress or promote it over ST006. It may remain useful as structure/capacity/source evidence.
- If compatible with your representation, prefer warm-starting / transferring from ST006 rather than restarting from O(1)-O(10) residual fields.
- If not compatible (e.g. source-faithful Kokuno mathematics), keep the route separate but explicitly use ST006 as the numerical benchmark to beat before PDE promotion.
- Do not import ST006 coefficients as source truth, reuse validation data for fitting, change forcing, or relax the fixed `1e-3` gate.
- Every future residual report must state whether it is directly comparable with ST006.

Coordination PR #290 now writes this rule into `PROJECT_GOAL.md`, `CURRENT_CHECKPOINT.md`, `AGENT_TASKS.md`, and `SCHEDULED_AGENT_TASKS.md`.

PR #300, opened 2026-09-18
Kokuno: bind source outer reserved-patch scales

Liang-techh:
## Task
`KOKUNO-A1-OUTER-RESERVED-PATCH-SCALES-011` — one minimal Kokuno Agent-1 leading-profile increment.

## Fresh audit / non-duplication
Before implementation I reread main/live README, AGENTS, PROJECT_GOAL, CURRENT_CHECKPOINT and AGENT_TASKS; audited open Kokuno PRs, recent scientific-branch commits and CI. This branch stacks on Agent-5 #295 exact head `4b206974e9912dbd31ddcb2b4dfe279ff49389f9`, so corrected heat V2 (#289), Agent-4 independent robustness (#294), and the current routing frontier (#295) are ancestry. It does not duplicate Agent-2 pulse/curl, Agent-3 mean/covariance or Agent-4 validation work.

## Source discovery that changes the blocker
The pinned corrected reader `KokunoYumeto/yang-mills-interacting-workbench@143f6773feb424ad9ed3a8d116653200f20346b7` gives more of the missing outer scale binding explicitly than the previous Agent-1 checkpoint recorded:

- `T_d=exp(M_d)+10`, `T_w=60 log(1/lambda)`;
- `X_R=110(C P_*)^10`, `X_w=X_R exp(T_d+2)`;
- reserved patches `I1..I4`, with `I2=X_w exp((T_w-20,T_w-15))` reserved for the heat compensation;
- exact reserved-patch base `U=0`, `E=c_patch (1+eta^2)^(-1) X^(-1/2-lambda)`;
- `P1=P_* exp(-1/5)`, `e_w=P1 exp(-T_d/2) exp(-1/2-lambda/2)`, `c_patch=e_w X_w^(1/2+lambda)`.

The source still does **not** publish one hidden numeric outer schedule. This PR therefore keeps the source free choices explicit and marks their concrete defaults as autonomous; it does not claim hidden-parameter recovery.

## One executable increment
Adds `KokunoOuterReservedPatchSchedule`:

- evaluates the source scale relations in log space, including all four reserved intervals;
- places `X_*` deterministically at the logarithmic centre of source `I2` (an explicit autonomous realization), while fail-closed checking that the complete existing #271 three-bump support stays inside `I2`;
- derives `e_*=c_patch X_*^(-1/2-lambda)` rather than guessing `X_*` and `e_*` independently;
[...]

PR #293, opened 2026-09-18
Kokuno A3: bind source cycle gains to real finite-cycle routing

Liang-techh:
## Task
`KOKUNO-A3-CORRECTION-CYCLE-GAIN-GUARD-009` — one minimal Kokuno Agent-3 correction-cycle/gain increment after #282.

## Fresh audit / non-duplication
Before implementation I reread the live `README.md`, `AGENTS.md`, `docs/PROJECT_GOAL.md`, `docs/CURRENT_CHECKPOINT.md`, and `docs/AGENT_TASKS.md`; audited live `codex/cr001-constraints@661cb9790b15f65b0d0eb97e9a457c8efe79ed0e`, current open PRs/recent commits and CI. Agent 1 #289 owns the corrected heat-tail target and still has no globally matched leading field. Agent 2 #291 now makes the projected pulse path inverse executable but explicitly does not instantiate the actual source pulse/background or materialize a second public `[u,v,w]` column. Agent 3 #282 still measures a 21.01945% transverse covariance remainder, with 25/27 active nodes requiring a genuinely second direction. Coordination #290 surfaces ST006 as the repository-wide numerical residual baseline.

This PR stacks directly on Agent-3 #282 exact head `56501232583f7e08f02c223e9b9e5b9d6ca94c27`. It does not construct an oscillatory column, duplicate Agent-2 curl/pulse code, alter Agent-1 leading work, fit pressure/forcing, or rerun a residual-selected parameter search.

## Why this increment
The corrected Kokuno reader's four-step correction-cycle audit gives source **exponent-class** gains with `kappa_s=1e-5`: wave residual `min(.4,.4-kappa_s,.5-2kappa_s,.5-2kappa_s)=0.39999`, full tangential mean `min(.17,1-4kappa_s,1-4kappa_s)=0.17`, and defect `0.9-4kappa_s=0.89996`; each exceeds the source-required `0.1`. Those are asymptotic class gains, not measured `residual_after/residual_before` ratios. The repository did not yet have an executable guard preventing those positive source margins from being misused to override a numerically divergent engineering correction cycle.

## One minimal executable increment
Adds `kokuno_correction_cycle_gain_guard.py` plus focused tests/workflow. It:
1. computes the source exponent gains from the displayed formulas rather than hardcoding a claimed contraction;
2. exposes an empirical exponent conversion only when the caller explicitly supplies a meaningful `Q` scale and declares source-class comparability;
[...]

PR #291, opened 2026-09-18
Kokuno A2: make projected pulse path inverse executable

Liang-techh:
## Task
`K2-OSC-010` — Kokuno Agent 2, one minimal projected-pulse increment after #281.

## Fresh audit / non-duplication
Before implementation I reread the latest README, AGENTS, PROJECT_GOAL, live CURRENT_CHECKPOINT/AGENT_TASKS, checked `codex/cr001-constraints@661cb9790b15f65b0d0eb97e9a457c8efe79ed0e`, open PRs, recent commits and current CI. Relevant newest lanes are Agent 1 #289 (corrected heat-tail factor; global leading still incomplete), Agent 3 #282 (measured missing second covariance-column target), Agent 4 #283 (independent heat audit), Agent 5 #284 (routing checkpoint), and coordination #290 (ST006 retained baseline/reporting rule). No current Agent 1/3/4/5 PR owns the source projected nonzero-harmonic pulse ODE.

This PR stacks directly on Agent-2 #281 exact head `4f9be3a159f02d0c3d4d30b07f847d02c3304f43` and does not change Agent-1 leading profiles or Agent-3 mean/radial correction.

## Pinned source equations
Provenance remains `KokunoYumeto/yang-mills-interacting-workbench@143f6773feb424ad9ed3a8d116653200f20346b7`, `navier-stokes/navier_stokes_workbench.tex`, corrected 2026-09-09 / Zenodo `22678406`.

The corrected reader records, for one nonzero harmonic,

`t_m' + K t_m + m^2 d t_m + i k_m n_Phi pi_m = -f_m`, `n_Phi dot t_m=0`,

`d = epsilon k^2 |n_Phi|^2`,

`pi_m = i/k_m * (n_Phi dot K t_m - n_Phi' dot t_m + n_Phi dot f_m)/|n_Phi|^2`,

and the equivalent projected equation

`t_m' = A_Phi t_m - m^2 d t_m - proj_(n_Phi^perp) f_m`,

`A_Phi = -K + n_Phi (n_Phi^T K - (n_Phi')^T)/|n_Phi|^2`.

The source proves `(n_Phi dot t_m)'=-m^2 d (n_Phi dot t_m)`, so zero entrance data remain transverse.

## One minimal executable increment
Adds `KokunoProjectedPulseInverseContract`:
- exact projected RHS, pressure recovery, diffusion and full-equation residual APIs;
[...]

PR #302, opened 2026-09-18
Kokuno A3: bound the missing second covariance inverse

Liang-techh:
## Task
`KOKUNO-A3-SECOND-COLUMN-BOUNDED-INVERSE-010` — one minimal Kokuno Agent-3 mean/covariance-correction increment after #293.

Before implementation I reread the live `README.md`, `AGENTS.md`, `docs/PROJECT_GOAL.md`, `docs/CURRENT_CHECKPOINT.md`, and `docs/AGENT_TASKS.md`; audited live `codex/cr001-constraints@e3611f84271fee1a9521b2908c4bd2fd762d58e5`, current open PRs/recent commits and CI. Agent 1 latest #300 binds a real source reserved-patch schedule/component but still does not materialize the final globally matched leading velocity. Agent 2 latest #291 makes the projected pulse path inverse executable but explicitly does not instantiate the actual source pulse/background or a second public `[u,v,w]` covariance column. Agent 3 #282 therefore still has a measured `21.0194502123%` transverse stress remainder, with 25/27 active nodes requiring a genuinely second direction; #293 still rejects the existing one-column finite cycle.

This PR stacks directly on #293 exact head `78c89b42f4ea5ad45884a292040339088111f9cf`. Final head is `7da46de54d8643b982e9f98e1e1a0eda082fbf14`. It does not construct/reimplement Agent-2 oscillatory curls, alter Agent-1 leading work, fit pressure/forcing, select residual parameters, or rerun the already rejected one-column PDE cycle.

## Why this increment
#282 can tell whether a future second response makes the local two-column map rank two and spans the real theta/axial compact-stress target. But in a two-component map, almost any nonparallel second column can algebraically span the target if its signed coefficient is allowed to become arbitrarily large. Kokuno's source construction instead uses an invertible two-column covariance map `H` together with a signed differential inverse. Before Agent 3 is allowed to materialize another public velocity correction, the repository needs an explicit scale/conditioning guard, not only a rank check.

For current response `c1`, future response `c2`, and #282's measured orthogonal remainder `m`, only

`c2_perp = c2 - proj_c1(c2)`

can remove `m`. At a rank-two node,

`|delta a_2| = |m| / |c2_perp|`,

and

`|det[c1,c2]| = |c1| |c2_perp|`.

Therefore a declared signed-coefficient budget `B` implies necessary local floors

`|c2_perp| >= |m|/B`,

`|det H| >= |c1||m|/B`.

These are engineering consequences of the signed inverse, not source numerical constants and not measured NS residual contraction claims.

Adds `kokuno_second_column_bounded_inverse.py`, focused tests and a dedicated report workflow. It:
1. consumes #282's real Agent-1 + Agent-2 defect/stress target; `surrogate_defect_used=false`;
2. derives the transverse-response and determinant floor required by a declared signed-update budget;
3. defines a data-bound symmetric budget as the largest signed coefficient already required by the existing rank-one target projection, plus routed amplitude `.125` only as a broader engineering reference scale;
4. evaluates an arbitrary future second covariance response for rank, condition number, exact stress coverage, both signed coefficients, and budget compliance;
5. calibrates duplicate, raw orthogonal-remainder, and scaled orthogonal-remainder cases without treating the abstract remainder as velocity;
6. refuses to authorize another finite correction cycle until an actual second public covariance column exists and passes this bounded inverse preflight.

## Exact-head real report
Inputs remain the real #282 path: leading candidate SHA256 `936768138f959e969caa5a8537647022f17818e65304ba7f60200d067c9bff67`, Agent-2 amplitude `.125`, phase `0`, `t=.5`, `z=.08`, annulus `[.08,.36]`.

Measured target:
- target vector RMS `1.0048446997713491e-05`;
- missing vector RMS `2.112128313791157e-06`;
- missing relative vector RMS `0.2101945021227428`;
- missing theta/axial RMS `1.9127752684896183e-06 / 8.957548694663862e-07`;
- 25 nodes require a second direction;
- max pointwise current/missing cosine magnitude `1.1703668490420364e-15`.

The existing rank-one solve requires coefficient RMS `0.006875958430184888` and max `0.012513055889932317`. Using that max as the symmetric engineering budget gives the following necessary future-column envelope on the 25 second-needed nodes:
- transverse response RMS `1.7541576728383513e-04`, max `3.1077227206743023e-04`;
- transverse/current median `0.08420532454169367`, max `0.2158081769179278`;
- determinant floor median `1.4826691746624834e-07`, max `4.4890101201743265e-07`.

Using the much broader routed primary-amplitude scale `.125` gives transverse/current median/max `0.00842932745776082 / 0.021603358234227456` and transverse response RMS/max `1.755989839998392e-05 / 3.110968647536805e-05`. This is reported only as an engineering reference, not a source smallness theorem.

Calibration exposes why rank alone is insufficient:
- duplicate current column: `0/25` rank-two required nodes; stress residual remains `0.21019450212274277`; preflight fails;
- raw abstract orthogonal remainder: `25/25` rank two and algebraic stress relative RMS `2.7719188534366146e-16`, but max second signed coefficient is `1.0000000000000009`, condition median/max `949.0674125649798 / 58154.29820191601`, so it fails the symmetric budget;
- the same abstract remainder rescaled to the symmetric budget: `25/25` rank two, both signed coefficients within budget, algebraic stress relative RMS `2.530825666676301e-16`, condition median/max `11.875733576739046 / 727.6879836403655`, so the algebraic bounded-inverse calibration passes. This scaled remainder is **not** candidate velocity and is not used as evidence of PDE improvement.

The actionable interface for Agent 2 is therefore quantitative: a future physical second public column is not enough merely to be nonparallel; at the tighter measured-update scale it must provide local transverse covariance derivative as large as about `21.58%` of the current response at the worst required node, with the corresponding determinant floor, or the signed inverse becomes too large before any PDE cycle is attempted.

## CI / reproducibility
Final exact-head dedicated run `35300304572`: **SUCCESS**. Focused command

`python -m pytest -q -W error tests/test_constrained_kokuno_missing_covariance_column_target.py tests/test_constrained_kokuno_second_column_bounded_inverse.py`

returned **9 passed in 0.64s**. Real report command

`python -m openai_ns_reconstruction.kokuno_second_column_bounded_inverse --output artifacts/kokuno_agent3/second_column_bounded_inverse_report.json --target-output artifacts/kokuno_agent3/second_column_bounded_inverse_target.json`

completed successfully. Artifact ID `10530095369`, digest `sha256:aa3e81c3adf09adc18df7e6b39d4d66efc7c6360421b2898a93f48dc2590b749`.

Standard exact-head tests run `35300304569`: **SUCCESS**; constrained lane `python -m pytest -q -W error tests/test_constrained_*.py` returned **234 passed in 132.18s**, and velocity/coordinates/forcing/import-CLI jobs also passed. Agent-5 reference-integration checkpoint run `35300304568`: **SUCCESS**.

The first dedicated run `35300184180` failed one synthetic assertion only because an algebraically exact coefficient landed a few ulps above the equality budget boundary. The final commit adds a fixed `1e-12` relative + machine-epsilon comparison allowance for that boundary. No candidate, source formula, coefficient budget, rank threshold, algebraic stress tolerance, amplitude, defect sample, PDE threshold, or acceptance criterion was changed to turn CI green.

## ST006 / truth boundary
ST006 remains the repository-wide numerical benchmark (momentum max `0.1082289305112118`, volume-L2 `0.10758432876230622`), but this report is a local covariance/stress inverse-scale diagnostic and is explicitly `directly_comparable=false` to ST006's full-domain normalized momentum protocol.

No new second public column exists in this PR, no velocity correction is materialized, no finite correction cycle is rerun, and no residual reduction is claimed. The next admissible Agent-3 action is to consume a genuinely independent Agent-2 public second column, run this bounded-inverse preflight on its measured covariance response, and only if it passes materialize the two-column signed velocity correction and run the frozen held-in/held-out residual-before/after cycle.

Hard states remain: `second_public_covariance_column_available=false`, `public_velocity_correction_materialized=false`, `finite_correction_cycle_run=false`, `residual_reduction_claimed=false`, `formal_full_domain_normalized_pde_gate_assessed=false`, `pde_validated=false`, `paper_exact=false`, `openai_field_identified=false`, `blowup_proved=false`.


Liang-techh:
Agent-5 #305 records #302's bounded-inverse preflight as ready but keeps `second_public_covariance_column_realized=false` and `correction_ready=false`. Do not rerun the rejected one-column cycle. Wait for Agent 2 to materialize a physical second public covariance velocity column; then apply #302's measured scale/conditioning guards before constructing the two-column signed velocity correction and running a frozen finite correction cycle.

PR #301, opened 2026-09-18
Kokuno A2: differentiate projected pulse into curl sensitivities

Liang-techh:
## Task
`K2-OSC-011` — Kokuno Agent 2, one minimal source-pulse sensitivity increment after #291.

## Fresh audit / non-duplication
Before implementation I reread the latest README, AGENTS, PROJECT_GOAL, live CURRENT_CHECKPOINT/AGENT_TASKS/SCHEDULED_AGENT_TASKS and checked `codex/cr001-constraints@e3611f84271fee1a9521b2908c4bd2fd762d58e5`, open PRs, recent coordination and current lanes. Relevant newest work is Agent 1 #289, Agent 3 #293, Agent 4 #294, Agent 5 #295, and coordination #290. Agent 5 explicitly identifies an independent source-motivated Agent-2 covariance/oscillatory column as the next missing seam; no other active lane owns projected-pulse parameter sensitivities or `D_r C_m/D_z C_m` algebra.

This PR stacks directly on Agent-2 #291 exact head `2655d90d9fa561eb49e982245ed550cf635271df` and does not change Agent-1 leading profiles or Agent-3 mean/radial correction.

## Source vs repository-derived formulas
Pinned source provenance remains `KokunoYumeto/yang-mills-interacting-workbench@143f6773feb424ad9ed3a8d116653200f20346b7`, `navier-stokes/navier_stokes_workbench.tex`, corrected 2026-09-09 / Zenodo `22678406`.

The reader-sourced projected pulse equation is reused from #291:

` t' = M t + g`, with `M=A_Phi-m^2 d I`, `g=-P f`, `P=I-n n^T/|n|^2`.

This PR then **analytically differentiates that sourced equation inside this repository** with respect to one caller-declared parameter `xi`:

` s'=M s + M_xi t + g_xi`, `s=partial_xi t`,

with differentiated transverse constraint

`n dot s + n_xi dot t = 0`.

For #281's source coefficient `C_m=i(n x t)/(k_m |n|^2)`, it likewise implements the repository-derived exact derivative

`C_m_xi = i/k_m * ((n_xi x t + n x s)/q - (n x t) q_xi/q^2)`.

The metadata explicitly distinguishes the reader-sourced parent equation from these analytic repository derivations; no sensitivity formula is presented as a paper quote.

## One minimal executable increment
Adds `KokunoProjectedPulseSensitivityContract`:
- analytic derivatives of `P`, `A_Phi`, `d`, `M`, and projected forcing;
[...]

PR #303, opened 2026-09-18
Kokuno A4: independently audit outer reserved-patch velocity

Liang-techh:
## Task
`KOKUNO-A4-OUTER-RESERVED-PATCH-INDEPENDENT-011` — one minimal Kokuno Agent-4 independent-validation increment after Agent 1 #300.

Before implementation I reread current main/live README, AGENTS, PROJECT_GOAL, CURRENT_CHECKPOINT, AGENT_TASKS and scheduled routing; audited open PRs, recent commits/CI and current Agents 1–3. Agent 1 #300 is the newest executable local leading-profile increment: source reserved-patch scales plus a domain-restricted pure-swirl `patch_velocity`, but no terminal-tail binding, applied heat compensation or global leading candidate. Agent 2 #301 only differentiates the projected pulse path and still has no actual public source oscillatory column. Agent 3 #302 only preflights the missing second covariance inverse and has no new correction velocity. Therefore a formal full-domain three-stage NS gate still cannot honestly be run.

This PR stacks directly on Agent 1 #300 exact head `8aab23c07e0d239624d7c72bb9c3e6577674d984` and does not modify its construction. Final Agent-4 branch head: `dc1ae2a111d0ea49c75d42fdd191d87b0abb989c`.

## Independent path
The production #300 path computes source scales in binary float and inverts native `q` with its contractive fixed-point solver. This audit instead:
- reconstructs every outer logarithmic scale from raw public parameters with 70-digit `Decimal` arithmetic;
- inverts `q-z^2 q^(2h)=1-t` by independent monotone bisection;
- evaluates the source power-law swirl formula directly and compares it only with public `patch_velocity(x,y,z,t)`;
- uses fresh seed `9173061`, three source-parameter cases and 12 off-grid points/case across `t=.375/.5/.625`;
- checks the fixed radial scale law and local solenoidality with independent Cartesian centered differences at relative steps `8e-4/4e-4/2e-4`;
- calibrates the divergence detector by adding a fixed 2% radial mutation to the public swirl field.

Predeclared local guards were encoded before CI: velocity relative agreement `2e-10`, log-scale absolute agreement `2e-11`, scale-law relative agreement `2e-10`, independent q-relation residual `2e-12`, finest local normalized divergence `2e-5`, and radial-mutation normalized divergence at least `5e-4`. These are implementation-consistency guards, not replacements for the formal project gates.

## Actual numerical result
All three parameter cases passed the local structural preflight.

Worst values across the 3-case matrix:
- public vs independent log-scale absolute error: `5.684341886080802e-14`;
- public `patch_velocity` vs independent source formula relative error: `2.906223819770527e-14`;
- independent q-relation relative residual: `3.2905291008925873e-16`;
- radial source scale-law relative error: `2.4735631987930877e-14`;
- local normalized divergence max at relative FD `2e-4`: `8.906451077397434e-08`.

The divergence result refines cleanly at approximately second order: worst max across the parameter cases is about `1.42522e-6 -> 3.56300e-7 -> 8.90645e-8` for relative steps `8e-4 -> 4e-4 -> 2e-4`.

Mutation calibration is strong: the fixed 2% radial mutation produces finest normalized divergence minima of about `1.60e-3`, `2.00e-3`, and `2.40e-3` for lambda `.04/.05/.06`, while the unmutated finest max stays below `9e-8`.

### Important nontriviality caveat
The local patch is finite and mathematically nonzero, but this is **not** the formal project nontriviality/energy gate. Under the autonomous source-scale choices, physical patch speeds are extremely small: the default case is about `6.75e-33 .. 1.24e-32`, and the three-case matrix spans roughly `4.35e-37 .. 9.46e-30`. That scale must be re-audited after the missing heat compensation/global assembly/physical scaling is materialized; this PR does not treat mere `>0` as final nontriviality acceptance.

## CI / reproducibility
Dedicated workflow run `35301062354` completed **SUCCESS** on the PR merge of exact base #300 + exact Agent-4 head. Focused suite: `10 passed in 0.48s`. Report artifact: `kokuno-agent4-outer-reserved-patch-independent`, artifact ID `10529956826`, digest `sha256:dbfe22c6df90fbecfb608d35dd09c8ed2b05b172ddaca251539562532b0bfe3a`.

Standard repository workflow `35301062359` also completed **SUCCESS**: constrained integration `246 passed in 31.35s`; velocity, forcing, coordinates and import/CLI slices all passed.

## Fixed truth boundary
The project full-domain references remain normalized momentum `<=1e-3` and divergence max `<=1e-5`; both remain **unassessed** here because #300 is only a local I2 component and there is no globally matched leading+oscillatory+correction `velocity/pressure/forcing` candidate. No training loss/tensor, candidate internal derivative, pressure/forcing fit or free `f=R` is used. ST006 remains the repository benchmark but is explicitly not directly comparable to this local source-map audit.

Result: `structural_preflight_passed=true`, but `formal_full_domain_pde_gate_assessed=false`, `pde_validated=false`, `paper_exact=false`, `openai_field_identified=false`.

Routing feedback: Agent 1 can treat the local reserved-patch scale/velocity mapping as independently implementation-consistent and move to terminal-tail binding + applying corrected V2 through the three-bump heat compensation/global leading assembly. When that global `velocity/pressure` candidate exists, Agent 4 returns immediately to the fixed held-out three-stage NS gate.


Liang-techh:
Agent-5 #305 now consumes #303 as the independent validation boundary for the I2 reserved patch. The local source-map/solenoidality seam is considered implementation-consistent and should not be repeated: Agent-5 rerun preserved max public/source velocity relative error `2.9062e-14` and finest local normalized divergence `8.90645e-8`. Please hold the formal validator for the first globally matched leading+oscillatory+correction `velocity/pressure/forcing` composite; only then run the unchanged held-out normalized momentum <=1e-3 and divergence <=1e-5 gate plus a directly comparable repository-baseline comparison.

PR #314, opened 2026-09-18
Kokuno A4: independently audit terminal-tail schedule

Liang-techh:
## Task
`KOKUNO-A4-TERMINAL-TAIL-INDEPENDENT-012` — one minimal Kokuno Agent-4 independent-validation increment after Agent-1 #310.

## Fresh audit / non-duplication
Before implementation I reread the latest main/live `README.md`, `AGENTS.md`, `docs/PROJECT_GOAL.md`, `docs/CURRENT_CHECKPOINT.md`, `docs/AGENT_TASKS.md`, scheduled routing, open PRs, recent integration commits and CI. The live scientific branch is `codex/cr001-constraints@7bf33488ce3531e606b8fdcfc4ea104c655b2831`; retained ST006 remains the repository PDE baseline with held-out momentum max `0.1082289305112118`, volume-L2 `0.10758432876230622`, and failed divergence / `pde_validated=false`.

Newest Kokuno lanes are Agent 1 #310 (terminal release/tail scales, but still no applied heat compensation or global leading field), Agent 2 #312 (source-normalized `D_z` pulse sensitivity, but no actual public second `[u,v,w]` column), and Agent 3 #313 (public covariance-tangent preflight, with the existing amplitude tangent still rank deficient and no finite correction cycle authorized). Therefore a formal global `leading -> +oscillatory -> corrected` NS gate still cannot honestly be run this round.

This PR stacks directly on Agent-1 #310 exact head `3ffff01961840b4fcecf1fe278af5b1cc7301789`. It does not modify Agent-1 construction, Agent-2 oscillatory algebra, Agent-3 correction routing, pressure, forcing, or any scientific threshold.

## Independent numerical path
Production #310 advances the two release transitions with `scipy.solve_ivp(method="DOP853")` and evaluates terminal integrals with Gauss-Legendre quadrature. Agent 4 instead uses:

- the exact scalar integrating-factor representation for `Q'+(1+l)Q=-l-h`;
- a uniform grid with cumulative **trapezoidal** integration only for the integrating-factor exponent;
- independent composite **Simpson** integration for transition forcing and all terminal integrals;
- three fixed resolutions `2049 / 4097 / 8193`;
- three explicit parameter cases: the default extreme schedule, the known moderate materializable schedule, and a perturbed outer schedule;
- terminal `Q(y)` probes at `0,.5,1,1.5,2,2.5,3`;
- direct reconstruction of the entire log-scale chain from raw schedule parameters rather than consuming production derived log properties;
[...]

PR #310, opened 2026-09-18
Kokuno: bind terminal release and heat-tail scales

Liang-techh:
## Task
`KOKUNO-A1-TERMINAL-TAIL-SCHEDULE-012` — one minimal Kokuno Agent-1 leading-profile increment, stacked directly on #300.

## Fresh audit / non-duplication
Before implementation I reread the latest repository/live README, AGENTS, PROJECT_GOAL, CURRENT_CHECKPOINT and AGENT_TASKS; audited open Kokuno PRs, recent commits and CI. Agent 2–5 currently own pulse/curl, mean/covariance, independent validation and integration/routing respectively. This PR stays in Agent-1’s leading/base-profile lane.

## Source formulas made executable
Pinned public source: `KokunoYumeto/yang-mills-interacting-workbench@143f6773feb424ad9ed3a8d116653200f20346b7`, corrected 2026-09-09 reconstruction.

The public outer schedule continues after the reserved power stage with:

- `X_p = X_w exp(T_w)` and pulse logarithmic length `13/lambda`;
- from the pulse endpoint, a fixed sufficiently large interpolation of length `T_f`, followed by an eta-independent hold of length `30 log(1/lambda)`; its endpoint is `X_rel`;
- exact release datum `Q_s(X_rel)=(lambda-h)/(1-lambda)`;
- `l:-lambda→-1` over one unit, `l=-1` for `4 log(1/h)`, then `l:-1→-h` over one unit;
- terminal multiplier `f_o(y)=1-rho_o psi_o(y)`, `rho_o=c_o h`, with `0<=f_o'/f_o<h/4`;
- `Q_p=∫_0^3 exp(∫_0^v(1+l)) f_o'/f_o dv`, terminal `l=-h+f_o'/f_o`;
- an intervening `l=-h` hold of exact length `(1-h)^(-1) log(Q_in/Q_p)`;
- `X_tail` at the terminal interval start, `X_K=e^.2 X_tail`, `X_b=e^3 X_tail`;
- terminal power `E=c_inf X^(-A) f_o`, so continuity fixes `c_inf` once the preceding amplitude chain is known.

The source also gives the exact steep-hold reductions `E_end/E_start=h^6` and `(X E^2)_end/(X E^2)_start=h^8`; these are retained rather than estimated away.

## Executable increment
Adds `KokunoTerminalTailSchedule`, stacked on `KokunoOuterReservedPatchSchedule` from #300. It:

[...]

PR #289, opened 2026-09-18
Kokuno: correct audited Delta C_p heat-tail factor

Liang-techh:
## Task
`KOKUNO-LP-HEAT-DISCREPANCY-TAIL-FIX-009` — Kokuno Agent 1, one minimal executable correction after Agent-4 #283 independently audited Agent-1 #280.

## Fresh audit / non-duplication
Reread latest main plus live `codex/cr001-constraints` README/AGENTS/PROJECT_GOAL/CURRENT_CHECKPOINT/AGENT_TASKS, open PRs, recent heads and CI routing. Agent 2 owns complete-curl algebra (#281), Agent 3 covariance target (#282), Agent 4 the independent heat-discrepancy evaluator (#283), and Agent 5 the routing block (#284). This PR stacks directly on #283 exact head `ecf1be298e880afe62b1f568a8f42bd27bc54d4a`; it does **not** reimplement the independent evaluator or add another outer/repair basis.

## Executable correction
Adds `KokunoHeatReplacementDiscrepancyV2`, preserving #280/v1 as historical evidence. For `y>=3`, where `chi_K=f_o=1`, write `H(z)-1=zR(z)`, `z=2d/X`, `A=1/2+h`. The source first-order `Delta C_p` tail is

`2*c_inf^2*d * int X^(-(2A+1)) R(2d/X) d(log X)`.

V1 carried one copy; V2 adds exactly the missing second copy using the existing production Gauss-Laguerre path. `Delta S` and `Delta I_sub` are unchanged. The v2 object has a distinct schema/SHA and fail-closed serialization so corrected and pre-audit targets cannot be confused.

## Independent cross-check
New tests consume #283's independent small-z/adaptive-log-space evaluator as an oracle rather than duplicating it. They require the v2-v1 change to be confined to `Delta C_p`, equal one half of #283's independently integrated full first-order tail, and require all three corrected channels to agree with the independent discrepancy across multiple eta probes. The deterministic `eta=.2` corrected checkpoint is approximately `[-2.21476226138e-09, 6.25423009649e-06, -2.63159129494]` within the independent tolerances.

## Truth boundary
This only repairs the source moment target. It does not recover `X_star/e_star`, instantiate the full outer schedule, apply the three-bump repair, create a global leading velocity, or assess the full NS gate. `pde_validated=false`, `paper_exact=false`, `global_leading_profile_reconstructed=false` remain unchanged.

## Exact-head verification
Final head: `eb736740373792f51b261549e958420d4b37b843`.

[...]

PR #271, opened 2026-09-17
Kokuno: add executable heat-discrepancy repair primitive

Liang-techh:
## Task
`KOKUNO-LP-HEAT-DISCREPANCY-REPAIR-007` — Kokuno Agent 1.

## Fresh audit / non-duplication
Before choosing scope I reread latest main and live scientific `README.md`, `AGENTS.md`, `docs/PROJECT_GOAL.md`, `docs/CURRENT_CHECKPOINT.md`, `docs/AGENT_TASKS.md`, checked the current live integration head, all current Kokuno PRs, recent commits and CI state. Agent 2 #262 owns source oscillatory phase/covector work; Agent 3 #263 owns signed complete-curl mean/covariance correction; Agent 4 #264 owns independent validation; Agent 5 #265/#268 owns routing/integration. None owns the source heat-replacement three-row compensation patch. #268 still points back to Agent-1 global/axial leading repair and matched core-to-heat assembly as the dominant scientific blocker.

This PR stacks directly on Agent-1 #261 exact head `fb53f6d984577386ac8beb02499690d099ded55d` and changes only the next source leading-profile primitive.

## Pinned public source / formula map
Provenance remains pinned to `KokunoYumeto/yang-mills-interacting-workbench@143f6773feb424ad9ed3a8d116653200f20346b7`, `navier-stokes/navier_stokes_workbench.tex`, corrected 2026-09-09 / Zenodo `22678406`.

For the earlier reserved patch the source writes
`x_*=X/X_*`, `E=e_* f x_*^(-1/2-lambda)`, `U=0`, and
`delta E=e_* sum_i c_i(eta) beta_i(x_*)`.
The three exact discrepancies come from
`delta(E^2)/(2X)`, `-delta(E^2)/2`, and `sqrt(2X) delta E`.
After dividing by `e_*^2`, `X_*e_*^2`, `X_*^(3/2)e_*`, the exact linear weights are
`f x_*^(-3/2-lambda)`, `-f x_*^(-1/2-lambda)`, `sqrt(2)x_*^(1/2)`.
The first two rows retain their exact quadratic terms; the third is linear. The source proves the ordered compact-bump moment matrix is nonsingular by the distinct-power/Chebyshev argument.

## Minimal executable increment
Adds `KokunoHeatDiscrepancyRepair` with:
- exact vectorized three-row nonlinear normalized discrepancy map;
- direct source linear-weight API;
- analytic 3x3 coefficient Jacobian;
[...]

PR #320, opened 2026-09-18
Kokuno: add signed-log V2 heat repair target

Liang-techh:
## Task
`KOKUNO-A1-LOG-RESCALED-HEAT-TARGET-013` — one minimal Kokuno Agent-1 leading-profile increment, stacked directly on #310.

Before implementation I reread the latest README, live AGENTS / PROJECT_GOAL / CURRENT_CHECKPOINT / AGENT_TASKS, current open Kokuno PRs, recent commits, and CI. Agent 2 owns pulse/curl sensitivity; Agent 3 covariance/mean tangent; Agent 4 independent validation; Agent 5 routing/integration. #314 independently confirmed #310’s terminal-tail chain and #315 explicitly routes Agent 1 to the log/rescaled V2 heat-compensation blocker. This PR only addresses that blocker.

## Source/provenance
Pinned public source remains `KokunoYumeto/yang-mills-interacting-workbench@143f6773feb424ad9ed3a8d116653200f20346b7`, corrected 2026-09-09 reconstruction. It reuses:

- #289 corrected V2 heat replacement, including the audited factor 2 in the first-order far-tail `Delta C_p` term;
- #300 source-derived `X_* / e_*` reserved-patch normalization;
- #310 source terminal `X_tail / c_inf / rho_o` scale chain;
- #271’s normalized three-row repair contract.

## Executable increment
Adds `KokunoLogRescaledHeatRepairTarget`.

Instead of forming absolute enormous `X_tail` and then evaluating tiny physical discrepancies, it substitutes `X=X_tail exp(y)` analytically and factors every power of `X_tail`, `c_inf`, `X_*`, and `e_*` into logarithms before quadrature. The O(1) transition/tail integrals are then combined by signed log-sum-exp.

API:

- `signed_log_target(eta)` -> exact channel sign plus natural-log absolute value for the corrected normalized target;
- `materialize_target(eta)` -> float64 only when every nonzero channel is actually representable; otherwise fail closed rather than silently writing zero;
- `rescaled_target(eta)` -> O(1)-referenced mantissa plus the full per-channel signed-log data;
- `precision_report(eta)` -> explicit underflow channels and estimated cancellation precision required by the existing float64 three-bump inverse;
- SHA-bound JSON save/load with source/autonomous/truth metadata.

## Cross-check against the existing V2 evaluator
A materializable autonomous schedule (`M_d=.1, lambda=.65, h=.005`) is evaluated both ways: through #289’s physical-scale V2 evaluator and this algebraically factored signed-log path. The three normalized channels agree to the test tolerance without using the new implementation as its own oracle.

## New numerical blocker made explicit
For #310’s default existence-style schedule at `eta=.2`, the new evaluator retains all three nonzero channels even though float64 cannot:

- `log10 |target_Cp| ~ -664.7`;
- `log10 |target_S| ~ -464.7`;
- `log10 |target_I| ~ -243.3`;
- signs are `(+,-,+)`.

Thus `C_p` and `S` are nonzero but below float64’s minimum subnormal, while `I_sub` is still representable. The channel span is >400 decades, so feeding `[0,0,target_I]` to #271 would not constitute faithful source compensation. This PR therefore fails closed instead of claiming that heat compensation is complete.

This is stronger than the previous generic “X_K^-1 underflows” warning: the actual corrected normalized target is now executable and its precision hierarchy is quantified.

## Tests
Adds constrained regressions for:

- independent agreement with #289 V2 on a materializable source-aligned schedule;
- default signed-log preservation of nonzero underflowed `C_p/S` channels;
- >400-decade float64 cancellation barrier reporting;
- rescaled mantissa plus preserved channel logs;
- exact `eta=+-1` zero endpoints under `-W error`;
- vectorization;
- SHA/JSON round-trip and truth-metadata mutation rejection.

## Truth boundary
This PR does **not** yet apply a three-bump correction to the I2 velocity profile. It specifically establishes that the existing float64 coefficient representation is not sufficient for the default source existence schedule without a higher-precision or hierarchical repair representation. Therefore:

- `actual_heat_discrepancy_applied_to_patch=false`
- `float64_three_bump_inverse_certified=false`
- `heat_compensation_completed=false`
- `global_leading_profile_reconstructed=false`
- `formal_full_domain_pde_gate_assessed=false`
- `pde_validated=false`
- `paper_exact=false`

Next Agent-1 step: use this signed-log target to build a precision-preserving/hierarchical three-bump coefficient representation, then apply it to I2 before any global promotion.


Liang-techh:
Kokuno Agent 4 independent handoff after exact-head audit PR #323 (`ff81bf49bc45d3575bd43a52113562354f381b3e`). I independently re-evaluated the new signed-log V2 heat target with a different path: exact small-Z heat limit, adaptive Gauss-Kronrod transition quadrature, analytic power-law tail, and independent `math.fsum` signed-log combination across three schedules / three accuracy levels.

Result: local structural preflight PASS. Max public-vs-independent log error `9.860627869784366e-9`; medium->fine log change `5.684341886080802e-14`; materializable moderate-case relative error `9.850282307155463e-9`; largest tested `log Z=-296.08034535881586`; recreated old half-tail Cp bug moves the Cp log by at least `0.002433440903587325`.

Most important handoff: default eta=.2 independently gives signs `(+,-,+)` and `log10|target|=(-664.7002251422,-464.6878051424,-243.2664224492)`. `C_p` and `S` are genuinely nonzero but float64-underflow, while `I_sub` is representable. So please keep the next three-bump coefficient solve hierarchical / signed-log / rescaled; converting the default target to `[0,0,I_sub]` before the solve would erase two verified constraints. I would move forward from the target map itself rather than spending another round on its quadrature. Formal full-domain PDE gate remains unassessed / `pde_validated=false` until the repaired global leading velocity/pressure exists.

Liang-techh:
Agent-5 routing after #323/#324: the signed-log target is now independently preflighted and bound into a fail-closed checkpoint. Please do **not** revisit X_star/e_star, X_tail/c_inf/rho_o, or the signed-log target derivation. Shortest next Agent-1 deliverable is the precision-preserving/hierarchical three-bump coefficient representation, application of the full nonzero signed-log target to I2, then the first global matched leading `velocity/pressure` contract. Default eta=.2 retains nonzero C_p/S below float64 subnormal (>400-decade channel span), so `[0,0,I_sub]` is not a faithful repair target. Agent-5 PR #324 / head `381c0cf...`; dedicated CI `35308676135` SUCCESS.

PR #331, opened 2026-09-18
Kokuno A4: independently audit hierarchical heat repair precision

Liang-techh:
## Task
`KOKUNO-A4-HIERARCHICAL-HEAT-REPAIR-INDEPENDENT-014` — one minimal Agent-4 independent-validation increment after Agent-1 #328.

## Fresh audit / non-duplication
This round reread the latest README, AGENTS, PROJECT_GOAL, CURRENT_CHECKPOINT and AGENT_TASKS on live `codex/cr001-constraints@1120071fb7426043eb6eb8bf7581e28a10e47169`; audited open Kokuno PRs, recent integration commits and CI. Agent 1 #328 is now the newest executable leading-side artifact: it preserves the >400-decade signed-log target in Decimal and applies a hierarchical three-bump correction on I2, but explicitly only certifies the repository's frozen discrete moment map and does not claim continuous source moment compensation or a global leading field. Agent 2 #321 still lacks an actual-source public `velocity(x,y,z,t)` oscillatory correction. Agent 3 #329 still screens covariance directions and does not authorize a finite correction cycle. Therefore a formal global `leading -> +oscillatory -> corrected` NS gate still cannot honestly be run this round.

This branch stacks directly on Agent-1 #328 exact head `3afbf5a019e09f9a351a49e09803e6d6ac25fc61`. Agent-4 exact head is `749b2fdae1d76787e88448ed15fe44098cd9f723`. It does not modify Agent-1 construction, pressure, forcing, thresholds, or any Agent-2/3 lane.

## Independent numerical path
Agent-1 #328 forms its three-bump moment tensors with the construction-side float64 Gauss-Legendre map, converts those tensor coefficients to Decimal, then solves the extreme signed-log target with high-precision Newton.

Agent 4 instead consumes only the public hierarchical coefficient receipt from `KokunoHierarchicalHeatRepair.solve(eta)` and reconstructs the moment rows from scratch using:

- a separately coded autonomous C-infinity three-bump basis;
- per-support composite Simpson quadrature at fixed `1024 / 2048 / 4096` interval resolutions;
- direct source linear weights for `C_p`, `S`, and `I_sub`;
- independently assembled quadratic bump-square terms for the first two rows;
- Decimal contraction only after those independent moment integrals are formed;
- four off-endpoint eta probes `-.6,-.2,.2,.6`;
- a fixed coefficient mutation (`c0 *= 1.001`) as a sensitivity calibration.

[...]

PR #330, opened 2026-09-18
Kokuno A2: execute source homogeneous primary pulse

Liang-techh:
## K2-OSC-015 — one bounded Agent-2 increment

Base: #321 exact head `1c5cf93b7468441a0f735336837a4ecdcb8e5afd`.

### Source/provenance
Pinned to `KokunoYumeto/yang-mills-interacting-workbench@143f6773feb424ad9ed3a8d116653200f20346b7`, corrected 2026-09-09 reader / Zenodo `22678406`.

The corrected reader distinguishes the zero-entrance forced inverse from its real **homogeneous primary pulse**. For the primary `m=1`, `f=0` pulse it specifies the moving transverse plane `U=[e_r-s_a K_a,N_a]`, eigenbasis `M`, `B=UM`, envelope `P(v)=exp(int a_net(|s|))`, entrance coordinates `z_+(0)=P(0), z_-(0)=0`, and a real `t cos(k Phi)` wave whose `m=+1` Fourier coefficient is `t/2`.

### Implementation
Adds `KokunoSourceHomogeneousPrimaryPulse`:
- implements the source `s(v)` and `a_net` formulas;
- evaluates `P(v)` through an exact elementary antiderivative derived from the displayed source integral, avoiding an additional quadrature approximation;
- constructs the source moving transverse plane and `B=UM`, with an explicit `n_Phi`-transversality guard;
- materializes the nonzero source entrance datum `t(0)=B(0)(P(0),0)`;
- reuses the existing guarded projected-pulse RK4 path solver with exactly `f=0` for the homogeneous pulse;
- returns the real cosine amplitude, the exact Fourier convention `t_+=t/2`, and the already verified complete-curl coefficient `C_+`;
- records deterministic provenance/truth-boundary metadata.

### Source vs autonomous choices
The moving-plane/envelope/entrance/Fourier formulas are source formulas. The closed-form antiderivative is an exact algebraic evaluation of the source integral. The existing tabulated-path RK4/interpolation and end-step orthogonal projection remain autonomous numerical approximations. Default pulse constants are bounded replay values and are **not** claimed to recover a hidden/source label.

### Deliberate limits
[...]

PR #329, opened 2026-09-18
Kokuno A3: screen second covariance columns across held-out times

Liang-techh:
## Task
`KOKUNO-A3-MULTISLICE-COVARIANCE-PREFLIGHT-013` — one minimal Agent-3 mean/radial-correction routing increment after #322.

## Fresh audit / non-duplication
Before implementation I reread the latest README, AGENTS, PROJECT_GOAL, CURRENT_CHECKPOINT and AGENT_TASKS; audited live `codex/cr001-constraints@1120071fb7426043eb6eb8bf7581e28a10e47169`, current open PRs, recent commits and CI. Agent 1 #320 remains on signed-log heat repair/global leading assembly. Agent 2 #321 materializes one source-real conjugate complete-curl pair from supplied mode data, but still explicitly lacks actual source pulse/background instantiation, direct public `velocity(x,y,z,t)` correction and a genuinely independent second source pair. Agent 3 #322 can screen one physical velocity column at `t=.5,z=.08`; its duplicate-column negative control remains rank deficient and another finite cycle is blocked.

This PR stacks directly on #322 exact head `2168b4fe8a7b2c6a177e6f9c4fd9a10428f61f95`. It does not copy Agent-2 pulse/curl code, change Agent-1 leading construction, fit pressure/forcing, or rerun the rejected one-column finite cycle.

## Why this increment
#302/#322 currently authorize or reject a future second covariance direction on one real defect/stress slice. But Agent-3's frozen finite-cycle validation already uses held-out times `.375/.5/.625`. A second column that only spans the `t=.5` stress target should not authorize a correction cycle.

This increment therefore reuses the same real defect -> compact theta/axial stress -> covariance remainder chain at all three frozen held-out times, while keeping the signed-coefficient budget frozen from the existing `t=.5` #302 receipt. No threshold is changed.

## Minimal executable increment
Adds `kokuno_multislice_covariance_preflight.py`:
- reconstructs real phase-mean theta/e=2 and axial/e=1 compact radial stresses at `t=.375,.5,.625`, `z=.08`, annulus `[.08,.36]`;
- measures the existing covariance response at each slice and builds the same pointwise rank-one remainder contract;
- freezes one coefficient budget from the existing `t=.5` real rank-one projection;
- evaluates any caller-supplied public phase-dependent velocity column through #322's multiplicative covariance adapter;
- routes every measured slice through the unchanged #302 rank/conditioning/signed-budget guard;
[...]

PR #332, opened 2026-09-18
Kokuno A5: bind hierarchical heat precision rejection

Liang-techh:
## Task
`KOKUNO-A5-HIERARCHICAL-HEAT-AUDIT-ROUTING-CHECKPOINT-014` — one minimal Agent-5 integration/artifact/CI increment after Agent-1 #328 and independent Agent-4 #331.

This round reread current README, live AGENTS / PROJECT_GOAL / CURRENT_CHECKPOINT / AGENT_TASKS, live integration PR #1, open Kokuno PRs, recent commits/Actions, and newest Agent 1–4 deliveries. It stacks directly on Agent-4 #331 exact head `749b2fdae1d76787e88448ed15fe44098cd9f723`; #331 itself stacks on Agent-1 #328 exact head `3afbf5a019e09f9a351a49e09803e6d6ac25fc61`.

Newest lane state consumed/routed here:
- Agent 1 #328: high-precision Decimal hierarchical solve of the signed-log I2 three-bump target. It closes the **repository frozen discrete** moment tensor to `<1e-40`, but explicitly does not certify continuous source moments or a global leading field.
- Agent 4 #331: independent per-support Simpson reconstruction rejects promotion of that discrete closure. Exact-head summary: max independent `I_sub` relative error `5.5244996273308184e-05`, minimum `C_p` log10(relative error) `416.9590367407875`, minimum `S` log10(relative error) `216.93128101171473`; the preregistered `I_sub <=1e-10` guard fails. This is a scientific rejection preserved by green CI, not a reason to weaken the guard.
- Agent 2 #330: source homogeneous primary pulse is now executable, but still requires caller-supplied source path geometry and does not expose a direct actual-source `velocity(x,y,z,t)` or an independent second covariance column.
- Agent 3 #329: multi-slice second-column preflight spans `t=.375,.5,.625`; 75 node/time requirements need rank coverage and the duplicate control achieves `0/75` rank two, so the finite cycle remains blocked.

## One minimal integration increment
Adds a fail-closed Agent-5 routing checkpoint, tests and artifact workflow. One command:

```bash
python -m openai_ns_reconstruction.kokuno_hierarchical_heat_audit_routing_checkpoint \
  --output-dir artifacts/kokuno_agent5/hierarchical_heat_audit_routing_checkpoint_v14
```

The checkpoint reruns/binds Agent-4's independent audit, reruns Agent-1's public hierarchical solve at `eta=.2`, serializes the construction-vs-independent precision gap, keeps ST006 as a non-source full-domain benchmark, and binds sibling Agent-2/3 routing without laundering those branches into executable ancestry.

## Routing result
The shortest leading-side blocker has changed again. **High-precision coefficient arithmetic is no longer the blocker; the underlying moment tensor / integration accuracy is.** The current coefficients close the frozen float64 moment tensor to `<1e-40`, while the independently integrated dominant `I_sub` row misses by `5.52e-5`, about 5.74 orders above its frozen `1e-10` independent guard; the tiny `C_p/S` rows miss their targets by hundreds of decades.

Therefore Agent 1 should not repeat Decimal Newton on the same float64 tensor. The next construction increment must rebuild the three-bump moment integrals/tensor at precision commensurate with the signed-log hierarchy, or use a cancellation-stable hierarchical formulation that avoids float64 tensor coefficients, then hand the result back to Agent 4's alternate-quadrature audit. Only after that audit passes should repaired I2 be promoted into a global leading `velocity/pressure` assembly.

Agent 2 should next bind actual source path/background geometry and source sensitivities into its pulse/complete-curl path, expose direct public Q-scaled `velocity(x,y,z,t)`, and provide a genuinely independent second covariance column. Agent 3 should route that column through the existing three-time guard before any new finite correction cycle.

## Exact-head verification
Final head: `9fc32ca30bcf9813b672ba977a7ced6ccd4fbbed`.

- Agent-5 dedicated routing workflow `35312853883`: **SUCCESS**. Focused Agent-1/4/5 hierarchical routing tests passed; bundle generation/receipt/upload all passed.
- Inherited Agent-4 independent workflow `35312853939`: **SUCCESS** on the same exact head.
- Standard pull-request test workflow `35312853768`: **SUCCESS**; velocity, coordinates, forcing, import/CLI and constrained-integration jobs all passed.
- Generated artifact: `kokuno-agent5-hierarchical-heat-audit-routing-checkpoint-v14`, artifact ID `10533722069`, digest `sha256:680184854e06953277ef67311141136b8fa2c8556a43f7469939890cf433f84b`.

Green CI deliberately preserves the **scientific rejection**: it does not turn the failed independent `I_sub <=1e-10` closure guard into success.

## Fixed truth boundary
No candidate, pressure, forcing, threshold or source identity is changed. ST006 remains the repository full-domain benchmark (`0.1082289305112118` max / `0.10758432876230622` volume-L2) and is not directly comparable with this local heat-moment audit. Fixed project gates remain normalized momentum `<=1e-3` and divergence `<=1e-5`. `leading_ready=false`, `correction_ready=false`, `velocity_export_ready=false`, `formal_full_domain_pde_gate_assessed=false`, and `pde_validated=false`. No free `f=R`, paper-exact, OpenAI-field or blow-up claim is introduced.


Liang-techh:
Routing update from Agent 1: #337 exact head `887fe6939a3d6fd09b40a090b671136a6d2e8b13` implements the #332 blocker directly. The three-bump I2 moment tensor is rebuilt with arbitrary-precision per-support tanh-sinh quadrature instead of promoting float64 Gauss-Legendre coefficients to Decimal. Dedicated CI `35315916416` and standard tests `35315916378` are green; the inherited independent 4096-Simpson dominant `I_sub` guard is now `<1e-10` (and >1000x better than legacy #328). Keep global promotion blocked until Agent 4 supplies an independent high-precision audit of the tiny `C_p/S` channels.

PR #336, opened 2026-09-18
Kokuno A2: differentiate homogeneous pulse spatially

Liang-techh:
K2-OSC-016. One minimal Agent-2 increment stacked on #330.

What changed:
- add an executable `D_r`/`D_z` spatial sensitivity contract for the source real homogeneous primary pulse;
- use the corrected-reader band-rectangle identity `D_r v=D_z v=0`, so the pulse coordinate/envelope are fixed under these spatial derivatives;
- analytically differentiate the moving transverse plane and nonzero entrance datum `t(0)=B(0)[P(0),0]^T`;
- propagate the nonzero entrance sensitivity through the existing projected-pulse sensitivity ODE and expose `t_plus`, `t_plus_xi`, `C_plus`, `C_plus_xi`;
- keep forcing identically zero and do not free-fit velocity components.

Truth boundary / limitations:
- caller still supplies `n_Phi,n_Phi',K` and the selected source-normalized spatial derivatives;
- actual positive-order background, auxiliary-torus derivative path, source labels/support partition, direct `velocity(x,y,z,t)`, full-domain PDE gate remain absent;
- not paper-exact and not identified as the OpenAI field;
- ST006 comparison is intentionally not made because this is a local derivative/structure contract, not the same full-domain PDE protocol.

Verification in the dedicated workflow checks the analytic moving-plane derivative against an independent fourth-order finite difference, re-solves the full homogeneous pulse at perturbed geometry to check `t_plus_xi/C_plus_xi`, checks refinement and differentiated transversality, and fail-closed provenance/serialization.


Liang-techh:
Agent-5 routing update from #340: your homogeneous-pulse spatial sensitivity lane is now recorded as ready/CI-green but remains sibling provenance only. Shortest next A2 deliverable is still the actual positive-order/source-background path and derivatives -> direct Q-scaled public `velocity(x,y,z,t)` -> a genuinely independent second covariance velocity column. Do not spend the next increment on another manufactured/projected duplicate column; Agent 3's spacetime admission guard is ready for the real one.

Liang-techh:
Agent 4 handoff from #348: the newly unified Agent-1 public reference router still has a stable O(10^1) raw momentum blocker under independent FD2 (finest h=.001 off-grid L2 32.1548/max 54.364; axis-near L2 12.9133/max 25.6499, with axis-near Rz RMS 12.8984). This makes a directly callable actual-source oscillatory `velocity(x,y,z,t)` more important than further local algebra alone: Agent 4 cannot honestly quantify leading -> +oscillatory contraction through the unified public contract until that evaluator exists. Please preserve an independent second covariance direction when materializing it; no free forcing / no gate change.

PR #368, opened 2026-09-18
Kokuno: make RF40 outer base schedule executable

Liang-techh:
## Scope
Kokuno Agent 1 minimal increment, stacked directly on #345 exact head `57df8375f134277ebea509847f6b319aa84acfd7`.

This implements the explicit corrected-reader RF40b/RF40c outer/base radial schedule as executable code. It does **not** route the result as the final segmented candidate across regions where the later cone or reserved-patch overlays are still missing.

## Pinned public source
`KokunoYumeto/yang-mills-interacting-workbench@143f6773feb424ad9ed3a8d116653200f20346b7`, `navier-stokes/navier_stokes_workbench.tex`, corrected 2026-09-09 / Zenodo `22678406`.

The source states, with `x=X/X_R`, `f=(1+eta^2)^(-1)`, and the log coordinate reset at each stage:
- temporary reference: `U=4 eta`, `E=P_* f exp(y/10)` for `y<=0`;
- length 1: `U=4 eta`, `l=(3/5)(1-sigma(y))`;
- length `T_d`: `U=k(y) eta`, `l=0`, `k(y)=4[1-sigma(log(1+y)/M_d)]`;
- length 1: `U=0`, `l=-lambda sigma(y)`;
- then length `T_w`: `U=0`, `l=-lambda`;
with `l=d_y log(sqrt(2X)E)` and therefore `d_y log E=l-1/2`. The endpoint amplitudes agree with the already-implemented RF40c relations `P1=P_* exp(-1/5)` and `e_w=P1 exp(-T_d/2) exp(-1/2-lambda/2)`.

## Executable increment
Adds `KokunoOuterBaseSchedule` with:
- vectorized `profile_values_logX(log_X,eta)` and materialized `profile_values(X,eta)`;
- exact source stage/reset semantics and the source flat step;
- analytic first `X`/`eta` derivatives for `E,F,U,v0` plus `Pi_X=F^2`;
- incompressibility-derived `v0` using the prefix moment ratio `M=eta*B(X)`;
- source-stage `velocity(x,y,z,t)->[...,3]` in the existing Kokuno native coordinates;
- log-space interval report, deterministic SHA, fail-closed JSON roundtrip.

Only the flat-step primitive and the stage-2 prefix moment use fixed Gauss-Legendre quadrature. The source endpoint integral `int_0^1 sigma=1/2` is pinned exactly rather than numerically approximated.

## Deterministic default-scale checkpoint
For the inherited explicit autonomous schedule `M_d=3`, `lambda=.05`, `h=.005`, `C=2`, `log(P_*)=T_d+1`:
- `T_d = 30.085536923187668`;
- `log X_R = 322.48732140326854`;
- first RF40b turn ends at `log X = 323.48732140326854`;
- axial-shutdown stage ends at `log X = 353.5728583264562`;
- `log X_w = 354.5728583264562`;
- `T_w = 179.74393641323945`;
- explicit RF40c power-law base stage ends at `log X = 534.3167947396956`.

The already-used I2 interval `[514.3167947396956, 519.3167947396956]` lies strictly inside this power-law stage, and the new evaluator is regression-tested there against `KokunoOuterReservedPatchSchedule`'s independently existing power-background formula.

## Tests / exact-head CI
Final exact head: `9a8042c89b0d10b94a17df81605338b9aee0f465`.
GitHub Actions run `35329961966`: **completed / success**.
- `constrained-integration` -> success;
- `slice-velocity` -> success;
- `slice-forcing` -> success;
- `slice-coordinates` -> success;
- `import-and-cli-help` -> success;
- optional legacy full matrix -> skipped by existing workflow policy.

The new governed tests check RF40 endpoint amplitudes, exact agreement with the existing reserved-patch power background, analytic derivatives against independent centered differences, the prefix-moment ODE, native Cartesian velocity replay, domain refusal after the explicit `T_w` stage, and fail-closed serialization. The Actions interface used here does not expose a trustworthy pytest pass count, so none is invented.

## Truth boundary
This is the **base** outer backbone. The final source field still requires cone modulation plus the I1/I2/I3/I4 overlays and matched pressure. Accordingly `cone_modulation_completed=false`, all four overlay flags remain false, `global_leading_profile_reconstructed=false`, `formal_full_domain_pde_gate_assessed=false`, `pde_validated=false`, and `paper_exact=false`.

No forcing fit, zero-field scaling, image-derived parameter recovery, or threshold change is introduced. Source/formula replay checks are not labeled independent PDE validation.


Liang-techh:
Agent-5 #374 records #368 as the current verified leading-side backbone: outer/base schedule executable and I2 contained, but `leading_ready=false` remains hard false. Shortest A1 work is still cone modulation + full I1/I2/I3/I4 overlay composition + globally matched pressure/admissible fixed forcing. Please do not divert into the A2/A4 source-shell preflight; that lane is now isolated separately. Agent 5 will only instantiate the final candidate artifact once this leading composite and the oscillatory/correction pieces coexist in executable ancestry.

PR #371, opened 2026-09-18
Kokuno A4: independently audit source wave shell in Cartesian coordinates

Liang-techh:
## Task
`KOKUNO-A4-SOURCE-WAVE-SHELL-CARTESIAN-AUDIT-019` — one minimal Agent-4 independent-validation increment stacked directly on Agent 2 #369 exact head `6050ea435536d01600d90a9dc02ebbab27293b94` (which inherits #358's source-grid/source-wave-shell wrapper).

## Fresh audit / non-duplication
Before implementation I reread current README, AGENTS, PROJECT_GOAL, CURRENT_CHECKPOINT and AGENT_TASKS; checked live `codex/cr001-constraints`, open PRs, recent commits/CI, and the newest Agents 1–3 state. Agent 1 #368 now has an executable RF40 outer/base schedule but still no global leading profile or matched pressure. Agent 2 #369 has a source squared-partition-family contract but still lacks concrete source bumps/labels, actual positive-order background, and a public source-derived xyz,t oscillatory velocity. Agent 3 #370 extends the radial correction side-effect audit but still has no genuinely independent second public covariance column and keeps the finite correction cycle blocked.

Agent 5 #362 and Agent 2 #358 specifically route the next Agent-4 increment to the source-grid/source-wave-shell seam. Prior Agent-4 #360 cleared the generic manufactured complete-curl seam using Cartesian FD4, but its inherited `p=1.35` phase was not source-grid compatible for every epsilon case.

## Independent increment
This PR keeps the frozen local guards from #353/#360 but changes the tested domain to the source-compatible seam:
- source labels satisfy `p=j/k`, `k=ceil(epsilon^-1/2)`, with integer physical azimuthal mode `m*j`;
- every held-out probe and every FD4 stencil point is required to satisfy `X=R^2/(2 s_Q)>=4/Lambda`;
- fresh seed `9173101`, 36 probes split equally between source-shell interior and near-inner-shell-edge strata;
- three source parameter cases and fixed Cartesian FD4 steps `.001/.0005/.00025`;
- independently recoded source-form phase/covector and `C_m=i(n x t)/(k_m|n|^2)`;
- independent Cartesian vector potential -> FD4 curl, versus only the public `KokunoSourceWaveShellLocalizedCurl.localized_mode(...)` output;
- caller `D_r C_m/D_z C_m` data come from a separate centered-FD6 path, not the Cartesian validation operator;
[...]

PR #369, opened 2026-09-18
Kokuno A2: enforce slow squared-partition family

Liang-techh:
K2-OSC-020, stacked directly on #358 exact head `e24d918bb0cc16b1f1006710183058296fbf760f`.

## Minimal increment
Make one additional corrected-reader support fact executable **without reconstructing a numerical cutoff**. The source records the slow labels `beta=(ell,a)`, rectangle labels `gamma=(ell,a,sigma)`,

`eta_beta = chi_ell(q) * chi_{ell,a}`, `sum_beta eta_beta^2 = 1`,

with duplicated +/- signs assigned after the beta partition. `KokunoSourcePartitionFamilyContract` now validates a whole caller-supplied beta family at once, rather than only the per-label `|eta|<=1` guard already present in #351.

Differentiating the squared partition gives

`sum_beta eta_beta D_r eta_beta = 0`,
`sum_beta eta_beta D_z eta_beta = 0`.

Those two first-derivative closures are explicitly labeled **autonomous algebraic consequences**, not separately quoted Kokuno formulas. They are checked in the same source-normalized `D_r/D_z` convention required by #351's product-rule complete curl.

The contract also fails closed if callers try to count `gamma=(ell,a,sigma)` +/- rectangle signs as separate squared-partition entries, or duplicate the same `beta` label.

## Source vs autonomous boundary
Source: beta/gamma bookkeeping; dyadic squared partition; `eta_beta=chi_ell chi_{ell,a}`; product-grid squared partition mesh `S_*^-3`; overlapping dyadic supports imply `|ell-ell'|<=2`.

Autonomous: floating-point tolerances and the two differentiated closure checks derived algebraically from `sum eta^2=1`.

Still **not reconstructed**: concrete numerical `chi_ell`, `chi_{ell,a}`, actual `S_*`, actual beta labels, outer annulus, positive-order background, or public xyz-t oscillatory velocity. No free forcing, residual fit, Agent-1 leading duplication, Agent-3 mean/radial correction, threshold relaxation, paper-exact claim, or OpenAI-field claim is introduced.

This PR does not bypass current Agent-4/5 routing: concrete source partition/background binding remains blocked until #358 receives the independent source-grid + source-wave-shell Cartesian/divergence preflight.

## Verification
Exact head: `6050ea435536d01600d90a9dc02ebbab27293b94`.

[...]

PR #380, opened 2026-09-18
Kokuno A2: compose localized real-pair family

Liang-techh:
K2-OSC-021, stacked directly on #369 exact head `6050ea435536d01600d90a9dc02ebbab27293b94`.

## Minimal increment
Compose the already-executable Agent-2 source-facing pieces without crossing the current Agent-4/5 routing boundary:

1. validate the whole caller-supplied `beta=(ell,a)` squared-partition family using #369;
2. for each beta, localize the vector-potential coefficient before curl using #351's exact product rule;
3. construct the real `m=+/-1` harmonic pair by conjugation and check it against `2*Re(m=+1)`;
4. apply caller-supplied **per-label** `Q_beta^(-A)` scaling before summing overlapping dyadic labels;
5. rotate the summed cylindrical result to Cartesian coordinates.

The new `KokunoSourceLocalizedRealPairFamily` accepts explicit per-beta `Phi, n_Phi, t_plus, D_r C_plus, D_z C_plus` and partition values/derivatives. It does **not** create a public `velocity(x,y,z,t)` map.

Source-facing identities reused: `eta_beta=chi_ell chi_{ell,a}`, `sum eta_beta^2=1`; localization at vector-potential level; real waves from conjugate nonzero harmonics; `u_phys=Q^(-A) E u_*`, `A=1/2+h`; slow-label wave assembly.

Autonomous composition only: whole-family-first runtime ordering, floating-point reality guard, exact shape/input checks, and per-label supplied-Q bookkeeping.

Important distinction: source rectangle sign `sigma` is **not** the Fourier harmonic sign `m=+/-1`; sigma is not inserted as an extra partition axis.

## Hard limits retained
Still false: concrete source partition bumps, actual source beta labels, actual positive-order/background path, auxiliary-dependent actual D_r path, public xyz-t velocity correction, genuinely independent second covariance column, formal full-domain PDE gate, PDE validation, paper-exact, OpenAI-field identification.

This PR does not change Agent-4 #371/#374's current independent preflight verdict and does not bind concrete source partitions/background until the routed independent preflight passes. No free forcing, residual fit, visual-only optimization, Agent-1 leading duplication, or Agent-3 mean/radial work is introduced.

## Verification
Exact head: `43f021e904dbb09c9bc2aba1d9ea2954083937e1`.

Dedicated workflow `35335492487`: **SUCCESS**. Focused suite: **13 passed in 0.49s**, covering the new family contract plus inherited partition-family and support-localized-curl checks. New tests compare the vectorized family against an explicit per-beta `m=+/-1` localized-curl sum, verify per-label `Q_beta^(-A)` scaling and Cartesian rotation, enforce `2*Re(m=+1)`, retain whole-partition closure diagnostics, reject partition/sign/Q misuse, and enforce provenance/serialization truth boundaries.

Standard repository workflow `35335492324`: **SUCCESS**. `constrained-integration`, coordinates, velocity, forcing, and import/CLI all passed; optional full matrix remained skipped under existing repository policy.


Liang-techh:
Agent 4 routing update from PR #383 exact head `71c6525ef710cc7bc2bf3599777da3b59af4b900`: the distinct preregistered roundoff-isolation audit on the source-grid/source-wave-shell seam now passes every unchanged local guard on fresh seed `9173111` with Cartesian FD4 steps `.004/.002/.001`. Curl relative RMS is `7.0518518228e-08 -> 4.4087562490e-09 -> 2.7557882720e-10` (ratios `15.9951, 15.9982`); divergence RMS is `3.1443063930e-07 -> 1.9657600246e-08 -> 1.2267404806e-09` (ratios `15.9954, 16.0243`); finest divergence max is `3.5807825887e-09`. The exact same `2e-4` absolute guards and `>=2.5` refinement rule were retained; #371's earlier rejection is preserved rather than overwritten. Mutation still gives curl relative RMS `0.0245123` and divergence max `0.0570100`.

This supports treating #371's final `~1e-10` plateau as a floating-point/FD floor rather than a persistent source-shell complete-curl leak. It does **not** validate a public actual-source oscillatory `velocity(x,y,z,t)`. Agent 2 can now prioritize the still-missing concrete source partition/background/actual labels and expose the real Q-scaled public xyz,t column; Agent 3 should only screen the resulting genuine second covariance column. Formal `1e-3` NS gate remains unassessed.

Liang-techh:
Agent-5 routing update from A4 #383: the source-grid/source-wave-shell complete-curl seam is now independently cleared by a distinct preregistered FD4 ladder with the original guards unchanged (dedicated run `35336259814` SUCCESS; finest curl rel RMS `2.7558e-10`, divergence max `3.5808e-9`, divergence refinement `15.995x/16.024x`). Historical #371 remains a retained reject; it is not rewritten.

Shortest A2 closure is now **not** more shell/curl validation: consume this cleared seam and bind the missing concrete source partition values/derivatives + actual positive-order/background path into #380's localized real-pair family, then expose a Q-scaled public `velocity(x,y,z,t)` and a genuinely independent second covariance column for A3. Keep supplied/manufactured data clearly separate from actual source inputs; do not promote until those inputs are real and provenance-bound. Agent-5 PR #385 tracks this routing.

PR #391, opened 2026-09-18
Kokuno: compute finite-frequency five-moment discrepancy

Liang-techh:
## Scope
Kokuno Agent 1 minimal executable increment, stacked directly on #381 exact head `65889fa6c7be1c8cdd8eb1f6c3d9bd1b9cd13122`.

Fresh audit reread the current README / AGENTS / PROJECT_GOAL / CURRENT_CHECKPOINT / AGENT_TASKS, current Kokuno PRs and exact-head CI. Agent 2/3/4/5 continue to own oscillatory, covariance/mean-correction, independent-validation and routing lanes; this PR only advances Agent 1's leading/base-profile path.

## Pinned public source
`KokunoYumeto/yang-mills-interacting-workbench@143f6773feb424ad9ed3a8d116653200f20346b7`, `navier-stokes/navier_stokes_workbench.tex`, corrected 2026-09-09 / Zenodo `22678406`.

Appendix C inserts the radial shear modulation by

`E_N = E exp(A(X,eta,N log X)/N)`,
`U_N = U + B(X,eta,N log X)/N`,

with `X d_X = D_X + N d_phi` and exact source shear identities

`a_N = a_L - 2 D_X A/N`,
`b_N = exp(-A/N) [b_L + 2 D_X B/(N E)]`.

It then measures five prefix-moment changes `M,J,I,S,C_p`; their negative discrepancy is the target restored by the later PA.17 I1 inverse delivered in #381.

## Executable increment
Adds `KokunoRadialModulationDiscrepancy` and governed tests.

The new object:
- evaluates source-form `E_N,U_N,F_N` and analytic first `D_X/eta` derivatives on the executable RF40 power-law backbone;
- checks the exact finite-N source shear identities against direct derivatives;
- computes the full nonlinear five-moment discrepancy in PA.17 row order `(M,J,I,S,C_p)`, retaining the `u e`, `u^2`, and `e^2` terms;
- performs the discrepancy integration in `y=log X` and analytically normalizes by #381's `X_1,e_1` scales, avoiding huge absolute-X powers;
- exposes the negative discrepancy as a directly consumable #381 target and demonstrates a pointwise target -> PA.17 solve;
- supports vectorized eta evaluation plus SHA/JSON round-trip and fail-closed truth metadata.

## Autonomous diagnostic realization / truth boundary
[...]

PR #382, opened 2026-09-18
Kokuno A3: retain cross-label mean covariance

Liang-techh:
## Task
`KOKUNO-A3-FAMILY-COVARIANCE-CROSS-TERMS-018` — one minimal Agent-3 mean-correction increment stacked directly on #370.

## Fresh audit / non-duplication
Before implementation I reread the live README, AGENTS, PROJECT_GOAL, CURRENT_CHECKPOINT and AGENT_TASKS; checked open PRs/recent commits/CI; and inspected the current Agent-1/2 handoffs. The new Agent-2 #380 supplied-data family exposes `velocity_physical_cylindrical_by_beta` and the assembled total, but still explicitly lacks concrete source bumps/labels, actual positive-order background/path, public xyz-t correction, or a genuinely independent second covariance column. This PR does not reimplement Agent-2 pulse, partition, localization, real-pair, or complete-curl code.

## Why this increment
Once overlapping slow labels are assembled, mean covariance is quadratic in the *sum* of label velocities. Summing only diagonal/self covariances discards beta!=gamma interactions and can therefore give the wrong theta/z mean-stress response and wrong second-column rank. For `W=sum_beta w_beta`, Agent-3 repository algebra gives

`C=(<W_r W_theta>, <W_r W_z>)`

and for a label tangent `s_beta`

`d_beta C=(<s_beta,r W_theta + W_r s_beta,theta>, <s_beta,r W_z + W_r s_beta,z>)`.

For dimensionless label multipliers (`s_beta=w_beta`), `sum_beta d_beta C = 2 C` exactly. These product-rule/self-cross identities are repository algebra, not additional Kokuno source formulas.

## Minimal executable increment
Adds `kokuno_family_covariance_cross_terms.py`, focused tests, and a dedicated report workflow. The contract:
- consumes an Agent-2-style mapping with physical cylindrical velocity by beta plus declared total;
- fails closed unless the total equals the sum of by-beta contributions and labels stay fixed over phase samples;
- measures assembled theta/z covariance, diagonal/self covariance, the retained cross-label remainder, and the full per-label amplitude Jacobian;
- enforces the quadratic Euler identity for unit label-multiplier semantics;
- accepts explicit tangent-scale factors when coefficient units differ, records those scales, and deliberately marks the result `bounded_inverse_ready=false` until an actual source coefficient-unit mapping exists;
[...]

PR #393, opened 2026-09-18
Kokuno A3: bound multi-label covariance inverse

Liang-techh:
## Task
`KOKUNO-A3-FAMILY-BOUNDED-INVERSE-019` — one minimal Agent-3 correction-machinery increment stacked directly on #382 exact head `c80d3d0993c0cc62e2f24cab5d427c5fcb12138d`.

Final exact head: `bef9206d3aac61f2de8f0c479dfa74cc40938746`.

Before implementation I reread the latest README, AGENTS, PROJECT_GOAL, CURRENT_CHECKPOINT and AGENT_TASKS; checked live `codex/cr001-constraints@89f290930e632fdea5449eac755183bc7f2024be`, open PRs/recent commits/Actions, and Agent-1/2 interfaces. Agent 1 latest #391 still does not expose a global leading profile/matched pressure. Agent 2 latest #380 exposes a caller-supplied localized real-pair family by slow label but still explicitly lacks concrete source beta labels/bumps, actual positive-order background/path, public xyz-t correction, and a genuinely independent second covariance column. This PR does not reimplement Agent-2 pulse, partition, localization or complete-curl code.

## Why this increment
Agent-3 #302 only bounded one future second covariance column. Agent-3 #382 now exposes the full cross-term-aware per-label covariance Jacobian for an overlapping slow-label family. A future concrete Agent-2 family may therefore have more than one additional response column. Simply checking every label coefficient separately would permit budget laundering: many duplicated or nearly duplicated columns can split one large correction into individually small coefficients.

For existing response `c0`, additional family responses `J=[c1,...,cm]`, and real measured target stress, this PR solves the local algebraic preflight

`[c0 J] da ~= target`

with the minimum-l2 least-squares solution, but accepts only if rank/coverage and explicit coefficient units are valid and

`|da0| <= B0`, `sum_j |daj| <= B_family`.

For exactly one additional column with equal budgets, this reduces to #302's existing per-coefficient rule. The aggregate L1 guard is autonomous anti-budget-laundering engineering, not a Kokuno source formula or a PDE threshold.

## Minimal executable increment
Adds `kokuno_family_bounded_inverse.py`, focused tests, and a dedicated workflow. The contract:
- consumes a real `MissingCovarianceColumnTarget` plus an arbitrary `(radial_nodes, labels, 2)` family Jacobian;
- always retains the existing current response as column zero;
- measures per-node rank, singular conditioning, minimum-norm coefficients and full algebraic stress residual;
- requires an explicit declaration that coefficient-budget units match the supplied Jacobian units;
- applies the frozen current-response coefficient budget separately and an aggregate L1 budget to all additional labels;
- fails closed on unit mismatch, rank deficiency, algebraic target miss, current-budget excess or aggregate-family-budget excess;
- includes a regression where two duplicate transverse labels would each appear individually below budget but their aggregate update exceeds the frozen budget, and correctly rejects it.

## Real-defect calibration at exact head
The report deliberately does **not** fabricate the missing actual-source family. It reuses the existing public complete-curl family as one label, with the same physical-amplitude tangent mapping `1/0.125=8`, and directly consumes the real #282 missing-stress target (`surrogate_defect_used=false`). Results:
- real active target nodes: `27`;
- nodes requiring a genuinely independent second direction: `25`;
- frozen signed coefficient budget: `0.012513055889617838` (`budget_changed=false`);
- missing relative covariance/stress response remains `0.210194502136004` (~21.01945%);
- one-label family Jacobian calibration error versus the existing current response: `0.0`;
- cross covariance for this one-label control: `0.0`;
- rank-two coverage: `0/25` required nodes;
- smallest singular value on required nodes: `8.668862731878635e-21`;
- algebraic relative stress residual remains `0.210194502136004`;
- `family_bounded_inverse_preflight_passed=false` and `finite_correction_cycle_rerun_allowed=false`.

This is the intended negative control: the existing family direction is just the already-known covariance direction and cannot fill the measured transverse defect.

No threshold, candidate, forcing, pressure, oscillatory parameter, or signed-coefficient scale was changed.

## Exact-head CI
Final exact head `bef9206d3aac61f2de8f0c479dfa74cc40938746`:
- dedicated `kokuno-agent3-family-bounded-inverse` run `35341173639`: **SUCCESS**; focused suite `15 passed in 51.11s`; real duplicate-family report executed and artifact uploaded;
- artifact ID `10545117446`, SHA256 `4781b7bc0c5bb7fc493c2caae6f8bdda06781f98805330752f4a4da67dfd8feb`;
- standard `tests` run `35341173484`: **SUCCESS**; constrained-integration `278 passed in 237.62s`; velocity / coordinates / forcing / import+CLI lanes also passed;
- Agent-5 reference integration checkpoint run `35341173431`: **SUCCESS**.

The first dedicated run on the prior head failed only because the new synthetic test fixture used five radial nodes while the existing target builder contract requires at least nine. The final commit changed that synthetic fixture from 5 to 9 nodes (and used `np.isclose` for the exact 0.5 synthetic coefficient assertion). No candidate, formula, real defect, signed budget, rank tolerance, algebraic tolerance, PDE threshold or acceptance rule changed.

## Next handoff
Agent 2 still needs to provide a concrete source-motivated `by_beta` public velocity family plus explicit coefficient normalization. Agent 3 should then:
1. measure #382's full cross-term-aware covariance Jacobian on the actual source family;
2. establish coefficient/Jacobian unit compatibility;
3. run this multi-label bounded inverse under the already-frozen spacetime/radial guards;
4. only if that passes, materialize the actual velocity correction while retaining the previously measured radial `d_z sigma_1` channel;
5. then rerun the frozen held-in/held-out finite correction cycle and independent normalized NS residual gate.

## Truth boundary
`actual_source_multilabel_family_consumed=false`, `source_coefficient_unit_mapping_available=false`, `actual_source_family_bounded_inverse_assessed=false`, `new_oscillatory_column_constructed=false`, `public_velocity_correction_materialized=false`, `finite_correction_cycle_run=false`, `residual_reduction_claimed=false`, `formal_full_domain_normalized_pde_gate_assessed=false`, `pde_validated=false`, `paper_exact=false`, `openai_field_identified=false`, `blowup_proved=false`.

The repository-level held-out normalized NS residual target `<=1e-3` is **not** claimed or assessed by this local algebraic preflight.


Liang-techh:
Agent-5 routing update from #395: #393's multi-label bounded-inverse machinery remains the correct fail-closed admission layer, but the current real negative calibration is still `0/25` rank-two with missing relative stress `0.210194502136004`; finite correction remains blocked. Once A2 publishes the actual/autonomous-provenance multi-label Q-scaled xyz,t family plus coefficient units, consume that exact family through #382 cross-label covariance and #393 aggregate-budget inverse. Only a frozen rank/algebraic-residual/unit/budget/spacetime/radial PASS may authorize materializing the correction (including radial `d_z sigma_1`) and rerunning the finite cycle.

PR #400, opened 2026-09-18
Kokuno A2: bound physical family coefficient units

Liang-techh:
K2-OSC-023, stacked directly on #392 exact head `e2f73ccdc4eccaebf03fd20e3b3d9c3d9df53977`.

## Minimal increment
The source-compatible partition seam is independently cleared by Agent 4 #394, and Agent 3 #393 explicitly needs a concrete physical family with declared coefficient/Jacobian units before its multi-label bounded inverse can be interpreted without budget laundering. This PR does not revisit curl, partition, leading, or mean/radial machinery.

It adds `KokunoBoundedFamilyCoefficientCoordinates`, a two-coordinate bounded normalization on the already-Q-scaled physical `by_beta` velocity emitted by Agent-2 #380/#392:

`m_beta = 1 + delta_common + delta_band g_beta`,

where `g_beta` is the deterministic affine contrast of the slow band label `ell` over the active family. The aggregate autonomous guard is

`|delta_common| + |delta_band| <= B <= 1/2`,

so every label multiplier remains positive. Parameter count is exactly two regardless of beta count; there is no free coefficient per label. At zero update the exact tangent columns are

`d_common u = sum_beta w_beta`,
`d_band u = sum_beta g_beta w_beta`.

Both coordinates have explicit units: dimensionless fractional multipliers of the already-localized, real-pair, per-label `Q^(-A)` physical velocity. A one-band family fails closed on nonzero `delta_band` rather than pretending it supplies a second direction. No phase-copy parameter is added because a phase copy is not a new covariance direction in the existing Agent-3 audit.

## Source/autonomous boundary
The corrected reader supplies the slow-label structure `beta=(ell,a)` and the upstream physical localized family. The two coefficient coordinates, affine band contrast, and aggregate L1 bound are repository-autonomous engineering choices. This PR does **not** claim recovered Kokuno primary amplitudes or the source signed differential inverse. It therefore sets `repository_coefficient_unit_mapping_available=true` while retaining `source_coefficient_unit_mapping_available=false`.

Still false: actual positive-order/background binding, public actual-source `velocity(x,y,z,t)`, genuinely independent second covariance column, finite correction-cycle permission, formal full-domain PDE gate, PDE validation, paper exactness, OpenAI-field identification, blow-up proof.

## Exact-head verification
Exact head: `d4464acc3e7072f473746df9b02df823cdbcb8c4`.

[...]

PR #392, opened 2026-09-18
Kokuno A2: realize source-compatible slow partition

Liang-techh:
K2-OSC-022, stacked directly on #380 exact head `43f021e904dbb09c9bc2aba1d9ea2954083937e1`.

## Minimal increment
Agent 4 #383 / Agent 5 #385 cleared the source-grid/source-wave-shell local preflight, so this PR does not revisit curl algebra. It materializes one concrete **repository-autonomous** slow squared-partition realization that can feed #380 with actual `eta_beta`, `D_r eta_beta`, `D_z eta_beta`, and unique `beta=(ell,a)` arrays.

The corrected 2026-09-09 reader specifies:
- `sum_ell chi_ell(q)^2 = 1`;
- three-coordinate product-grid squared partitions;
- per-band mesh `S_*^(-3)=ell^(-6)` on the active high bands;
- `eta_beta=chi_ell*chi_{ell,a}` and `sum_beta eta_beta^2=1`;
- rectangle `sigma` signs are assigned after the beta partition;
- overlapping dyadic supports obey `|ell-ell'|<=2`.

The public reader does **not** specify one unique numerical smooth bump family or hidden grid origin. Therefore the concrete C-infinity bump `rho(s)=exp(-1/(1-s^2))`, the `y=-log2(q)` placement convention, first admissible integer band `ell_min=5`, grid origin, and finite label enumeration are explicitly autonomous choices. This PR does not call them recovered Kokuno data.

## Implementation
`KokunoSourceCompatiblePartitionRealization`:
1. builds an l2-normalized compact dyadic family;
2. for each active ell builds an independently l2-normalized 3-D tensor-product grid with the source mesh `ell^-6`;
3. forms `eta_(ell,a)=chi_ell chi_(ell,a)`;
4. differentiates the raw bumps and l2 normalization analytically for caller-supplied source-normalized `D_r/D_z q` and slow-coordinate derivatives;
5. passes the flattened family through the existing #369 whole-family partition/derivative guard;
6. emits arrays directly consumable by #380.

No free forcing, residual fit, Agent-1 leading/profile work, Agent-3 mean/radial correction, or visual optimization is introduced.

## Exact-head verification
Exact head: `e2f73ccdc4eccaebf03fd20e3b3d9c3d9df53977`.

[...]

PR #401, opened 2026-09-18
Kokuno A3: bridge bounded family coordinate units

Liang-techh:
## Task
`KOKUNO-A3-BOUNDED-COORDINATE-UNIT-BRIDGE-020` — one minimal Agent-3 correction-machinery increment stacked on Agent-3 #393 and explicitly consuming Agent-2 #400 as a dependency.

Final exact head: `c8db7fc707e74beef510cdd90da06b5cf2295682`.

Before implementation I reread current README, AGENTS, PROJECT_GOAL, CURRENT_CHECKPOINT and AGENT_TASKS; checked live integration, open PRs, recent commits and CI, and inspected current Agent-1/2 interfaces. Agent 2 #400 newly exposes exactly two bounded repository coefficient coordinates on its already-Q-scaled physical `by_beta` family: `delta_common` and `delta_band`, both dimensionless fractional multipliers. Agent 3 #393 already owns cross-term-aware covariance Jacobians plus the multi-label bounded inverse. This PR does not reimplement Agent-2 pulse, partition, localization, complete curl, or physical family construction.

## What this fixes
#393's frozen signed correction budget is measured in the existing **physical-amplitude coordinate**. #400's new coordinate unit is a **dimensionless fractional multiplier** of an already-physical family. Reusing the same numeric budget in #400 coordinates would silently change the physical correction scale.

For the calibrated routed family `W=a w_unit`, preserving the same physical velocity-update scale requires repository algebra

`B_fractional = B_amplitude / |a|`.

The exact real-target receipt gives:
- frozen physical-amplitude budget: `0.012513055889932317`;
- reference physical amplitude: `0.125`;
- converted fractional budget: `0.10010444711945854`;
- `budget_changed=false`.

This is a coordinate-unit conversion, not a threshold relaxation. The bridge fails closed if the converted budget exceeds #400's positivity-preserving `<=1/2` coordinate contract or if an explicit normalization is missing.

## Mean-covariance handoff
For either #400 velocity tangent `S_j`, the full assembled-family response is measured with all label cross interactions retained:

`d_j C_theta=<S_j,r W_theta + W_r S_j,theta>`,
`d_j C_z=<S_j,r W_z + W_r S_j,z>`.

For `delta_common`, `S_common=W`, so the exact quadratic identity `d_common C=2C` is enforced. On the real one-band routed complete-curl calibration:
- common-coordinate response relative error against `0.125 *` the old physical-amplitude response: `4.78894088298019e-16`;
- common quadratic-identity relative error: `4.78894088298019e-16`;
- band-coordinate covariance-response vector RMS: `0.0`;
- one-band `delta_band` is correctly inactive.

## Real defect / bounded inverse result
The report directly consumes the real Agent-1 + routed Agent-2 target (`surrogate_defect_used=false`) at `t=.5`, `z=.08`, annulus `[.08,.36]`.

The new unit-correct handoff does **not** manufacture an independent correction direction:
- active target nodes: `27`;
- nodes requiring a second direction: `25`;
- rank-two required nodes: `0/25`;
- minimum smallest singular value on required nodes: `0.0`;
- algebraic relative stress residual RMS remains `0.2101945021227428` (~21.01945%);
- `family_bounded_inverse_preflight_passed=false`;
- `finite_correction_cycle_rerun_allowed=false`.

So the intended negative control is preserved: `delta_common` is only the existing direction in new units, while one-band `delta_band` is exactly zero.

## Verification
Exact-head dedicated run `35346154634` — **SUCCESS**:
- focused Agent-3 tests: `10 passed in 32.69s`;
- real-defect report step: SUCCESS;
- report artifact ID `10547465914`;
- artifact digest `sha256:84298de509d06cd6bd11a9e701069c3e5c6a3578be0ee0faf3726f5b17450dd6`.

Repository workflow `35346154364` — **SUCCESS**:
- constrained integration: `283 passed in 284.76s`;
- import/CLI, velocity, forcing and coordinates lanes all succeeded.

Agent-5 reference integration checkpoint `35346154402` — **SUCCESS**.

Upstream Agent-2 #400 exact-head workflows `35345330212` and `35345330333` are also SUCCESS.

An earlier stale-head attempt failed only because I temporarily copied Agent-2 #400's own test/workflow without its #392-only localized-real-pair dependency into the Agent-3 stack. The fix removed those copied Agent-2 CI files while retaining the #400 bounded-coordinate module and explicit dependency provenance. No candidate, real defect, formula, correction budget, rank tolerance, residual threshold, or PDE acceptance condition was changed.

## Truth boundary / next dependency
The actual source multi-band positive-order/background family is still absent. #400's coefficient coordinates remain explicitly repository-autonomous: `source_coefficient_unit_mapping_available=false`. No source-independent second covariance direction has been produced, no public correction velocity has been materialized, and no finite correction cycle or full-domain PDE residual-before/after has been run.

Next Agent-3 action should occur only after Agent 2 supplies a real multi-band/source-motivated physical family with explicit normalization. Then this bridge can measure its `delta_band` covariance direction, route it through #393 plus the frozen spacetime/radial guards, retain the previously measured radial `d_z sigma_1` side effect, and only if those gates pass materialize `delta_u` and rerun the finite correction cycle.

No free `f=R`, no training-point-as-validation substitution, no threshold manipulation, no finite-stage blow-up claim, and no duplication of Agent-2 oscillatory curl work. Held-out normalized NS residual `<=1e-3` remains unmet.


Liang-techh:
Agent 4 routing from #402: Agent-2 #400's repository-autonomous two-coordinate value map independently clears a fresh multi-band black-box audit. Without reading #400 tangent arrays, held-out public modulated totals recover common/band covariance response with ~2x first-order refinement across `.04/.02/.01`; finest relative RMS is `5.0000e-3` (common) and `3.40885e-3` (band), common `dC=2C` error `1.34e-15`, label permutation error `1.06e-16`, and deliberate label/velocity misbinding changes the band response by relative `2.0`. Dedicated #402 run `35346814135` is SUCCESS (`9 passed`).

This supports using #400's **repository-autonomous** coordinate units as a numerically coherent handoff, but does not change #401's scientific blocker: the current real family is one-band, so `delta_band` remains inactive and there is still no genuinely independent actual-source second covariance column. Finite correction cycle remains disallowed until a real multi-band/source-motivated physical family exists and passes your existing unit/rank/radial-side-effect guards.

Liang-techh:
Agent-5 #403 has now bound this unit bridge into the integration frontier. The important result is preserved fail-closed: unit compatibility is solved, but the one-band calibration is still `0/25` rank-two with residual `0.2101945021227428`, so no correction/final-cycle promotion is allowed. Next Agent-3 action should wait for Agent-2's first real multi-band physical by-beta family, then run the existing #401 unit bridge + #393 bounded inverse under the frozen aggregate budget/spacetime/radial guards. Please keep `d_z sigma_1` and cross-label terms in the materialized correction path.

PR #430, opened 2026-09-18
Kokuno A2: execute source band covering schedule

Liang-techh:
## Task
`K2-OSC-026` — one minimal Agent-2 source-binding increment stacked directly on #420 exact head `1c77b07a0e8286b8ffa487917f48d54f6017c7e0`.

## Fresh audit / non-duplication
Reread current README, AGENTS, PROJECT_GOAL, CURRENT_CHECKPOINT and AGENT_TASKS; checked #420 CI (both dedicated and standard workflows are now completed/success) and latest Agent 1/3/4/5 handoffs. Agent 1 #409 still owns/incompletely assembles the leading field (inner-reference→RF40 join, I3/I4, global pressure remain blocked). Agent 3 #415 owns nonlinear covariance gain and still lacks a genuine second direction. Agent 4 #412 independently cleared the auxiliary-torus differential pullback. Agent 5 #413 routes Agent 2 toward source band/mode binding and public by-beta/total velocity. This PR does not duplicate those lanes.

## Source increment
The corrected 2026-09-09 reader displays a deterministic per-band schedule once `ell` and `h` are fixed:

- `Q=2^(-ell)`, `epsilon=Q^h`, `S_*=ell^2`;
- `i=floor(log_{T_g}(Q^(-1-h)/S_*))`;
- `c_i=T_g^i Q^(1+h)` and `M_i=Lambda_g^i Q^(d_r/2)`;
- exact floor bounds `T_g^-1 S_*^-1 < c_i <= S_*^-1` and `T_g^-rho_g epsilon^-kappa_s S_*^-rho_g < M_i <= epsilon^-kappa_s S_*^-rho_g`;
- product-grid mesh `S_*^-3=ell^-6`;
- for interacting bands within four dyadic levels, the displayed finite covering-level difference bound.

Adds `KokunoSourceBandCovering` to make those formulas, scale-free bound diagnostics, and interacting-band receipt executable. It reuses #420's displayed source constants and radial exponent rather than introducing a second copy.

The source only proves existence of suitable rational rectangle centers and sufficiently small `r0`; it does not publish one unique numerical tuple. `pulse_length_from_r0` therefore requires an explicit caller value and labels it `caller_supplied_not_source_recovered`. Rectangle centers, grid origin, and `r0` remain hard-unrecovered.

## Verification
Branch head at PR creation: `167d8e750d7ba5e4abaeaec7fc93ccf70145b7fd`.

[...]

PR #429, opened 2026-09-18
Kokuno: execute Appendix-B Xi boundary profiles

Liang-techh:
## Scope
Kokuno Agent 1 minimal executable increment stacked directly on #421 exact head `cac0fe6d6172b544f800d7330944c4ad074d2ea1`.

This round reread the current README/AGENTS/PROJECT_GOAL/CURRENT_CHECKPOINT/AGENT_TASKS, live constrained branch, recent CI/commits and open Kokuno PRs. Agent 2/3/4/5 remain on oscillatory/source-torus, covariance/correction, independent-audit and routing lanes. #424 still routes Agent 1 to the missing upstream Appendix-B boundary data before the later PA.16 join can be called source-backed.

## Pinned public source
Provenance remains `KokunoYumeto/yang-mills-interacting-workbench@143f6773feb424ad9ed3a8d116653200f20346b7`, `navier-stokes/navier_stokes_workbench.tex`, corrected 2026-09-09 / Zenodo `22678406`.

The public continuation gives the reference primitive equations

`D_X p1_r = X S_q,r/L - l_r p1_r`, `D_X n_s,r + n_s,r = S_n,r/L`,

then activates stress with

`e_a=(1-kappa_0)sigma(y/t1)`, `kappa=1-e_a`,
`a=kappa p1_r`, `D_X U=-kappa X n_s,r/2`, `D_X log(phi)=-kappa p1_r/2`.

After keeping `kappa=kappa_0` through `X=100`, the source shuts off the axial prescription on a short logarithmic interval, then keeps `D_X U=0` while interpolating `a` to `0.8`; the final interval has `a=0.8`, `l=0.6`, `D_XU=0` through `X_i=110`. The source leaves `kappa_0` and the two final widths as existence choices.

## Minimal executable increment
Adds `KokunoAppendixBBoundary`:
- reuses the executable public reference continuation rather than inventing a second natural profile;
- evolves the source reference primitive ODEs for `p1_r,n_s,r` and the source activation ODEs for the selected actual `F,U`;
- uses the source exact `S_q,S_n` formulas, including pressure and prefix-moment terms; after the reference freeze it propagates those terms analytically in `X` instead of repeatedly re-integrating the frozen profile;
- makes finite autonomous choices `kappa_0=0.01` and two `0.02` logarithmic final widths, records them as autonomous source-compatible existence choices, and checks that they fit strictly inside `100<X<110`;
[...]

PR #431, opened 2026-09-18
Kokuno A3: validate frozen damping on disjoint spacetime cells

Liang-techh:
## Task
`KOKUNO-A3-DAMPING-HOLDOUT-TRANSFER-024` — one minimal Agent-3 correction-machinery increment stacked directly on #422 exact head `96c250baedc6ad7af12acf954ed9166b045cb916`.

Reread current README, AGENTS, PROJECT_GOAL, CURRENT_CHECKPOINT and AGENT_TASKS; checked open PRs, recent commits/CI and current Agent-1/2 handoffs. Agent 1 #429 now executes selected Appendix-B Xi boundary profiles but still lacks incoming five-moment/T_sh closure plus global leading pressure/field. Agent 2 #430 (head `167d8e750d7ba5e4abaeaec7fc93ccf70145b7fd`, dedicated and standard CI SUCCESS) executes the public source band-covering schedule but still explicitly lacks the actual positive-order/background multi-band physical mode family and a genuinely independent public second covariance direction. This PR does not reimplement those lanes.

## Why this increment
#422 correctly forces one shared damping across its 3x3 spacetime screen, but that same 3x3 target set is used to select the shared `lambda`. Those cells are therefore **damping calibration data**, not independent validation data. The project contract explicitly forbids using the same fitted points as held-out validation.

This PR makes the split executable. It keeps #422's calibration grid exactly

`t=(.375,.5,.625)`, `z=(.06,.08,.10)`,

freezes the existing negative common-coordinate update and the shared quartic damping selected only there, then applies that exact fixed step with **no retuning** to a disjoint interleaved screen

`t=(.3125,.6875)`, `z=(.07,.09)`.

The fresh cells reconstruct the real Agent-1 + routed Agent-2 compact theta/axial stress target (`surrogate_defect_used=false`). The physical-amplitude/fractional budgets, quadratic-identity tolerance, algebraic tolerance, candidate, forcing and PDE thresholds are unchanged. The validation contract deliberately has no local-optimum field and rejects any time/z overlap with calibration.

## Exact-head numerical receipt
Exact head: `9d2c234234e46b53bf141d08d458ae206409ca4d`.

Calibration reproduces #422 exactly: shared negative-coordinate damping `lambda=0.4926349922838171`, calibration aggregate compact-stress relative residual `0.6223507487797298`.

With that scalar frozen **before** the fresh screen is evaluated, the disjoint 4-cell validation gives:
- fixed-damping aggregate compact-stress relative residual `0.6223757026222282`;
- full-step aggregate residual `0.9861371010802802`;
- zero-update reference `1.0`;
- `validation_transfer_preflight_passed=true`, which means the same frozen step improves over zero update and is no worse than the full step in **all 4/4 fresh validation cells**;
- `heldout_damping_reoptimized=false`, `validation_objective_used_for_selection=false`.

The calibration→validation shift in the aggregate local stress metric is only about `2.50e-5` absolute, while the validation grid was not used to choose `lambda`. This corrects the validation-separation weakness in #422 and shows that its approximately half-step negative update transfers to the predeclared disjoint spacetime cells. It is still a compact covariance/stress diagnostic, **not** a held-out NS residual reduction.

## Fail-closed routing
The fresh transfer result does not unlock the finite NS correction cycle. The current one-band family still fails the upstream bounded-inverse/rank gate (`upstream_bounded_inverse_passed=false`), no actual source multi-band family or genuinely independent second covariance column exists, the public correction is not materialized, and the previously measured radial `d_z sigma_1` side effect is not yet retained in a materialized correction. Therefore `finite_correction_cycle_rerun_allowed=false` remains hard false.

No free `f=R`, pressure/forcing fit, validation reuse, budget expansion, threshold relaxation, finite-stage blow-up claim, Agent-2 oscillatory duplication, or PDE promotion is introduced.

## Verification
All exact-head workflows are green:
- dedicated disjoint-transfer workflow `35364119783`: **SUCCESS**; focused suite = **13 passed in 0.90 s**; real calibration+fresh-validation report succeeded;
- report artifact `10555722075`, digest `sha256:cc069df53d630fbb66aae8af3a374fc9d61fe148071673d45fb66e763d612b00`;
- standard repository workflow `35364119580`: **SUCCESS**; constrained reconstruction lane = **300 passed in 277.77 s**, import/CLI, velocity, forcing and coordinates lanes all green;
- Agent-5 reference checkpoint `35364119732`: **SUCCESS**.

## Truth boundary
`damping_calibration_validation_disjoint=true`; `heldout_damping_reoptimized=false`; `validation_objective_used_for_selection=false`; `surrogate_defect_used=false`. Still hard false: `actual_source_multiband_family_consumed`, `genuinely_independent_second_covariance_column_ready`, `public_velocity_correction_materialized`, `radial_force_retained_in_materialized_correction`, `finite_correction_cycle_run`, `residual_reduction_claimed`, `formal_full_domain_normalized_pde_gate_assessed`, `pde_validated`, `paper_exact`, `openai_field_identified`, `blowup_proved`. Held-out normalized NS residual `<=1e-3` is not assessed by this local stress-transfer increment.

Next seam remains source-driven rather than threshold-driven: once Agent 2 exposes the actual multi-band physical family with an independent covariance direction, route it through the existing unit/rank/budget/spacetime/radial guards, then use the same calibration/fresh-validation separation for damping before materializing `delta u` and running a genuinely disjoint held-in/held-out finite NS correction cycle.


Liang-techh:
Agent 4 routing from #432: Agent 2 #430's deterministic source band-covering schedule has now passed an independent 80-digit Decimal audit, including 16 near-floor transition cases and 1245 interacting-band pairs, with mutation sensitivity. This removes the **schedule arithmetic** as a blocker, but it does **not** unlock the finite correction cycle. Your current fail-closed condition remains correct: no actual source multi-band physical family / genuinely independent second covariance column / public materialized velocity correction exists yet. Keep the held-out damping-transfer result separate from any full NS claim, and wait for the actual multi-band by-beta/total velocity family before rerunning rank-two/correction promotion. Formal `1e-3` momentum and `1e-5` divergence gates remain untouched.

Liang-techh:
Agent-5 #433 now stacks directly on this exact head and records the completed receipt: dedicated `35364119783`, standard `35364119580`, and Agent-5 reference `35364119732` all SUCCESS; artifact `10555722075`, digest `sha256:cc069df53d630fbb66aae8af3a374fc9d61fe148071673d45fb66e763d612b00`. The disjoint transfer result (`0.6223757026222282` fixed-damping vs `0.9861371010802802` full step) is retained as local stress validation only. Because rank remains `0/25` and upstream bounded inverse false, finite-cycle routing stays closed. Next A3 action should wait for A2's actual multi-band physical family, then run the unchanged unit/rank/budget/spacetime/radial/quadratic guards and repeat this disjoint transfer on the actual proposed correction.

PR #432, opened 2026-09-18
Kokuno A4: independently audit source band covering

Liang-techh:
## Task
`KOKUNO-A4-SOURCE-BAND-COVERING-INDEPENDENT-AUDIT-025` — one minimal Agent-4 independent-validation increment stacked directly on Agent 2 #430 exact head `167d8e750d7ba5e4abaeaec7fc93ccf70145b7fd`.

## Fresh audit / why this seam
Reread current README, AGENTS, PROJECT_GOAL, CURRENT_CHECKPOINT and AGENT_TASKS; checked current open PRs, recent commits/CI, and latest Agent 1 #429 / Agent 2 #430 / Agent 3 #431 handoffs. The complete global `leading + oscillatory + correction` candidate still does not exist, so the formal held-out NS gate cannot honestly be run this round.

Agent 2 #430 newly makes the corrected-reader deterministic band-covering schedule executable. Its own tests use ordinary float formulas close to the implementation path. This PR independently audits that seam with high-precision arithmetic before later source-family binding relies on it.

## Independent operator and frozen guards
The audit treats `KokunoSourceBandCovering` as a public value-only contract. Reference values are rebuilt from the displayed source constants with **80-digit Decimal arithmetic**, and the reference covering level is obtained by repeated high-precision powers rather than Agent 2's float-log/floor path. It does not call Agent-2 `bound_diagnostics()` or `interaction_receipt()`.

Before execution the local guards were frozen:
- public-vs-Decimal relative error `<=5e-12` for Q, epsilon, grid mesh, c_i and M_i;
- zero covering-level mismatches on the broad sweep;
- zero mismatches on 16 adversarial floor-transition cases placed about `1e-9` and `1e-12` to either side of integer covering levels;
- interacting-band bound formula relative error `<=5e-12` and zero violations over an exhaustive `ell=12..96`, `|delta ell|<=4`, three-h sweep;
- off-by-one covering-level mutation must move the weaker scale channel by relative `>=1`;
- wrong grid exponent `ell^-5` vs `ell^-6` must move the mesh by relative `>=4`.

Formal project gates are unchanged: normalized momentum max/L2 `<=1e-3`, divergence max/L2 `<=1e-5`.

## Exact numerical receipt
Exact branch head: `ada7e63fe8cd95acbb5c1b8bb2d646a1eefc644d`.

Dedicated run `35364815755` **SUCCESS**. Focused parent + independent suite: **8 passed in 0.51 s**. Generated audit says `local_structural_preflight_passed=true` with:
- 54 broad `(ell,h)` cases; worst value relative error `2.2898921537313096e-14`; covering-level mismatches `0`;
- 16 transition-near cases; minimum tested distance to an integer level `9.96933756555282e-13`; mismatches `0`;
- 1245 interacting-band pairs; violations `0`; maximum observed level difference `2`; tightest source-bound slack `1.0367263925746588`; independent bound-formula relative error `1.462394540831668e-16`;
- weakest off-by-one-level mutation relative change `1.5857864376269046`;
- wrong mesh-exponent mutation relative change `4.000000000000001`.

Uploaded report artifact ID `10556286420`; ZIP SHA-256 `eba4bb13efc2a9c19fe406621baaa681e6d921d83ee8d27592b1f404153dc5a3`.

Standard repository run `35364815707` **SUCCESS**. Governed constrained-integration suite: **291 passed in 132.91 s**; forcing, velocity, coordinates, and import/CLI lanes all completed successfully; the optional full matrix was skipped under repository policy.

## Truth boundary / routing
This is only a local source-schedule audit. It does not bind the still-missing actual positive-order/background, actual auxiliary-torus multi-band mode family, public xyz,t oscillatory velocity, global matched leading pressure, or correction composite. No training loss/tensor, pressure/forcing fit, free `f=R`, threshold relaxation, source-hidden parameter recovery, or PDE promotion is introduced.

The deterministic Q/epsilon/i/c_i/M_i/product-grid/interacting-level schedule is no longer an Agent-4 blocker. Agent 2 should move to the actual positive-order/background + source multi-band auxiliary-torus family and expose public Q-scaled by-beta and total `velocity(x,y,z,t)` with explicit coefficient units/normalization. Agent 3 remains fail-closed until that actual multi-band family provides a genuinely independent second covariance direction.

`formal_full_domain_pde_gate_assessed=false` and `pde_validated=false` remain hard false.


Liang-techh:
Agent-5 #433 refreshed after this receipt landed. v26 now records #432 as the latest Agent-4 handoff: exact head `ada7e63fe8cd95acbb5c1b8bb2d646a1eefc644d`, dedicated `35364815755` SUCCESS, standard `35364815707` SUCCESS, artifact `10556286420` / `sha256:eba4bb13efc2a9c19fe406621baaa681e6d921d83ee8d27592b1f404153dc5a3`. The source-band covering arithmetic is therefore removed from the independent-validator blocker list. Next highest-value A4 work remains the first actual multi-band physical family; no need to keep re-auditing Q/epsilon/i/c_i/M_i or product-grid floor behavior unless Agent 2 changes that interface.

Liang-techh:
Agent-5 v27 routing continues to treat #432's source band-covering audit as stable/cleared. No further Agent-4 time should go into Q/epsilon/covering-level arithmetic unless a new regression appears. The next high-value independent target is the first actual Agent-2 multi-band physical `by_beta`/total xyz,t family (including genuine second covariance direction); after the complete frozen composite exists, Agent 4 remains the owner of the unchanged full-domain `1e-3` momentum / `1e-5` divergence gate and same-protocol ST006 comparison.

PR #440, opened 2026-09-18
Kokuno A3: gate frozen damping by radial divergence

Liang-techh:
## Task
`KOKUNO-A3-RADIAL-AWARE-DAMPING-HOLDOUT-025` — one minimal Agent-3 correction-machinery increment stacked directly on #431 exact head `9d2c234234e46b53bf141d08d458ae206409ca4d`.

## Fresh audit / non-duplication
Reread current README, AGENTS, PROJECT_GOAL, CURRENT_CHECKPOINT and AGENT_TASKS; checked open PRs, recent commits/CI and current Agent-1/2 handoffs. Agent 1 #438 advances the selected Appendix-B five-moment target but still leaves source pressure-datum binding/global leading completion fail-closed. Agent 2 latest #439 certifies a source-compatible rational rectangle separation witness but still explicitly does not materialize the actual positive-order/background multi-band physical mode family or public xyz,t oscillatory correction. This PR does not reimplement either lane.

## Increment
#431 validates one frozen damping on spacetime cells disjoint from damping calibration, but its transfer objective contains only the theta/z compact covariance-stress channels. #357/#370 separately established the corrected-reader source requirement

`(div T)_r = + d_z sigma_1`.

For #415's already-fixed full velocity update `V`, the exact damped covariance change is

`Delta C(lambda)=lambda B(W,V)+lambda^2 C(V)`.

The axial component is the proposed `sigma_1` change. This PR differentiates that same frozen damped axial covariance stress on the unchanged `.02 -> .01 -> .005` z ladder, then applies the source `+d_z sigma_1` sign to the real ring-averaged gated radial phase-mean defect on #431's disjoint validation cells.

The routing rule is deliberately conservative and introduces no scientific threshold: a step that passes #431 theta/z transfer is still rejected if its signed radial divergence increases gated radial-defect RMS on any validation cell. Damping/sign/budgets are not reoptimized on validation data.

## Exact real-defect receipt
Exact head: `ea593e2f5ed128a4f31a8da3419755e49fa8ad3f`.

[...]

PR #439, opened 2026-09-18
Kokuno A2: certify rational rectangle separation

Liang-techh:
## Task
`K2-OSC-027` — one minimal Agent-2 source-compatible realization increment stacked directly on #430 exact head `167d8e750d7ba5e4abaeaec7fc93ccf70145b7fd`.

## Fresh audit / non-duplication
Reread the current README, AGENTS, PROJECT_GOAL, CURRENT_CHECKPOINT and AGENT_TASKS; checked open PRs, recent commits/CI and current Agent-1/3/4/5 handoffs. Agent 1 still owns the incomplete leading/background assembly. Agent 3 #431 keeps finite correction fail-closed until an actual multi-band family supplies a genuine second direction. Agent 4 #432 independently clears #430's deterministic band-covering arithmetic. Agent 5 #433 routes Agent 2 toward the actual multi-band physical family. This PR does not duplicate those lanes.

## Source increment
The corrected 2026-09-09 reader proves finite rational auxiliary-torus centers can be chosen so that

`c_mu != J_g^Delta c_nu (mod Z^2)`

for every finite relevant `0<=Delta<=Delta_max`, excluding only the tautology `Delta=0, mu=nu`. It then defines the positive finite separation `d_c`, `C_J=max ||J_g^Delta||`, `C_v=|v_r|+|v_t|`, and requires the strict guards

`4 r0 C_v < 1`,
`2 r0 C_v (C_J+1) < d_c`.

Adds `KokunoRationalRectangleSeparation` with exact `Fraction` torus arithmetic, exact transformed-collision checking, exact squared-separation storage, source `C_v/C_J` guards, and direct compatibility with #430's `L_s=2r0/c_i`.

The source proves existence but does **not** publish one unique center tuple, color count, denominator or `r0`. Therefore `autonomous_rational_witness` / `autonomous_for_band_window` construct a deterministic repository-autonomous rational witness and mark its center grid and safety fraction as source-compatible autonomous choices, not recovered source parameters. `Delta_max` can be chosen as the integer ceiling of #430's displayed four-band covering-level bound.

## Exact-head verification
Exact head: `e5bc36691d16be09cab602885fc49cb99f3d2f71`. Compare against #430 is ahead 3 / behind 0 with exactly three added files and no modification/deletion of inherited source.

[...]

PR #449, opened 2026-09-18
Kokuno A3: certify joint correction gain margin

Liang-techh:
## Task
`KOKUNO-A3-JOINT-GAIN-MARGIN-026` — one minimal Agent-3 correction-cycle gain increment stacked directly on #440 exact head `ea593e2f5ed128a4f31a8da3419755e49fa8ad3f`.

Reread current README, AGENTS, PROJECT_GOAL, CURRENT_CHECKPOINT and AGENT_TASKS; checked open PRs, recent commits/CI and current Agent-1/2 handoffs. Agent 1 #443 numerically instantiates the PA.10 `T_sh` certificate but still leaves source pressure datum/global leading completion fail-closed. Agent 2 #439 certifies rational rectangle separation and its dedicated + standard workflows are now SUCCESS, but it still does not materialize the actual positive-order/background multi-band public physical velocity family or a genuine second covariance direction. This PR does not reimplement either lane.

## Increment
#431 froze one scalar damping on calibration cells and transferred it without retuning to disjoint theta/z compact-stress cells. #440 additionally checked the corrected-reader radial side effect `(div T)_r=+d_z sigma_1` at that same frozen damping. This PR adds the finite-step gain diagnostic those routes still lacked.

For the already-fixed physical update `V`, the retained covariance change is

`Delta C(lambda)=lambda B(W,V)+lambda^2 C(V)`.

Every squared theta/z stress residual and squared radial residual proxy is therefore a quartic polynomial in `lambda`. On the unchanged disjoint #431/#440 validation cells this PR:

1. forms exact quartic coefficients from the real candidate defect/stress targets and measured linear/self covariance;
2. keeps the source `+d_z sigma_1` sign and decomposes its finest radial derivative into linear/quadratic pieces, with an independent identity check back to #440's direct damped stress;
3. solves each quartic sublevel set on `[0,1]`, extracts the connected admissible interval containing the already-frozen #431 damping, and intersects all 8 constraints;
4. reports left/right distance-to-failure margins and binding cells.

No validation cell chooses a new damping. Frozen damping/sign/budget/cells and all scientific thresholds are unchanged. Root imaginary/cluster tolerances (`1e-9`, `1e-10`) are explicitly recorded as numerical degree-4 root-classification tolerances only, not PDE/gain thresholds.

Tangential admissibility inherits #431's stronger per-cell rule: remain no worse than both zero update and that cell's full step. Radial admissibility inherits #440's threshold-free rule: source-sign radial gated-defect RMS remains no worse than zero update.

## Exact real-defect receipt
Exact head: `383eeb622e40bc4ac46f02a369bd205ca003adc8`.

Frozen inputs are unchanged: `lambda=0.4926349922838171`, `delta_common=-0.10010444711945854`, `delta_band=0`, fractional L1 budget `0.10010444711945854`, validation `t=(0.3125,0.6875)`, `z=(0.07,0.09)`, 33 radial nodes, 8 angles, 8 phases. `budget_changed=false`, `damping_changed=false`, `validation_used_for_damping_selection=false`, `surrogate_defect_used=false`.

All 8 disjoint-cell constraints contain the frozen damping. Their joint connected admissible interval is

`[0.02862512785452287, 0.9858531565488349]`.

Hence the already-frozen half-step sits strictly inside the joint interval with left margin `0.4640098644292942`, right margin `0.4932181642650178`, minimum two-sided margin `0.4640098644292942`, and total interval width `0.957228028694312`.

Binding constraints are tangential, not radial:
- left boundary: `z=.09` tangential cells at both fresh times; frozen relative stress RMS about `0.61835735`, inherited full-step bound about `0.96432263`;
- right boundary: `z=.07` tangential cells at both fresh times; frozen relative stress RMS about `0.62808518`; their admissible upper root is about `0.98585316`;
- all four radial source-sign constraints admit the whole `[0,1]` interval under their inherited non-worsening rule. At the frozen step they reproduce #440's relative radial proxies: about `0.99206330` at `z=.07` and `0.99319723` at `z=.09`.

Thus `joint_gain_margin_preflight_passed=true`. This establishes that the #431/#440 frozen damping is not a narrow single-point success on the current disjoint screen; it has a large finite perturbation margin under the inherited local compact-stress/radial gain rules.

## Truth boundary
This remains a robustness diagnostic around the existing one-band negative control, not a public velocity correction and not a full Navier--Stokes residual. The actual source multi-band family and genuine independent second covariance column are still absent; the upstream bounded inverse remains false. Therefore `finite_correction_cycle_rerun_allowed=false`, `finite_correction_cycle_run=false`, `residual_reduction_claimed=false`, `formal_full_domain_normalized_pde_gate_assessed=false`, and `pde_validated=false` remain hard false.

## Exact-head verification
All exact-head workflows are green:
- dedicated `kokuno-agent3-joint-gain-margin` run `35375639370`: **SUCCESS**; focused tests, real-defect report, and artifact upload succeeded;
- standard repository `tests` run `35375639385`: **SUCCESS**; constrained-integration plus import/CLI, velocity, forcing, and coordinates lanes succeeded;
- Agent-5 reference checkpoint `35375639420`: **SUCCESS**;
- report artifact id `10560506690`, digest `sha256:99d303fa79bf192921f1836cb1935235795beb253d6587fd5c2146925b3fcf85`.

Next required step remains Agent 2's actual source-motivated multi-band public physical velocity family with a genuine independent covariance direction. Only after the existing unit/rank/budget/spacetime/quadratic/radial/gain-margin guards pass should Agent 3 materialize a public `delta_u` and execute a fresh held-in/held-out NS cycle.


Liang-techh:
Agent 4 update from PR #451: the new Agent-2 #450 source-scheduled multiband family is numerically consistent with an independent Cartesian curl/divergence oracle on the frozen 36-point, 3-resolution screen, but the public handoff is **not yet structurally admissible** for your correction lane. Exact head `3e8170bde17cfeee5189e453f304388e65564931`, dedicated run `35377042248`: curl relative RMS `8.6017e-4 -> 5.9106e-5 -> 3.7898e-6`; normalized divergence RMS `1.1744e-3 -> 8.3970e-5 -> 5.45064e-6`; refinement ratios are all ~14–16. However a semantically identical simultaneous beta-data+label permutation is rejected by #450 at 12/36 held-out points because the post-Q^(-A) by-beta total and sorted by-band regrouping are compared with an absolute-only 2e-12 aggregation guard. Agent 4 therefore records `local_structural_preflight_passed=false` and has routed the order-invariance bug to Agent 2; no threshold was relaxed. Please keep `finite_correction_cycle_rerun_allowed=false` / no public correction promotion until Agent 2 fixes this and A4 reruns the same frozen invariant. Formal NS gates remain unassessed.

PR #516, opened 2026-09-18
Kokuno A5: route audited numeric prerequisite seams

Liang-techh:
## Kokuno Agent 5 — audited numeric prerequisite routing v35

One minimal integration increment stacked directly on Agent-5 #511 exact head `d244804a4b2322127f42239803d7023557be2325`. Final exact head: `02d3c2b5663434101d9517a82d5cabe0fb667df6`.

### Fresh audit / non-duplication
Reread current README, AGENTS, PROJECT_GOAL, CURRENT_CHECKPOINT and AGENT_TASKS; refreshed open Kokuno PRs, recent commits and Actions. Latest sibling deliveries routed here are Agent 1 #512, Agent 2 #513, Agent 3 #514, and Agent 4 #515. This PR does not rewrite their mathematical lanes and does not claim sibling PRs are executable ancestry.

### Minimal integration increment
Adds deterministic Agent-5 checkpoint v35 and fail-closed CI/tests for four now-executable prerequisites:

1. **Leading / A1 #512** — the coupled shared-C PA.10 necessary target band for `log E_i` is executable. The previously selected realization remains excluded. Passing this pointwise screen means only `not excluded`; it is not a source `B_0/T_sh` certificate and does not permit PA.16 by itself.
2. **Oscillatory mass / A2 #513** — the source-displayed `I0_sigma -> D_g -> h_sigma` mass law is numerically executable on explicitly provenance-labelled candidate pulse/cutoff samples over the independently audited autonomous geometry. Those samples remain candidate/repository choices; actual source pulses/background/modes and public physical `velocity_osc(x,y,z,t)` are still absent.
3. **Correction input / A3 #514** — a deterministic repository-autonomous finite-head factor is executable (`prepared_N=5`, `band=5`, `q=1.3`, `q_phys=0.040625`, factor `0.28410624176179694`). It is explicitly not the formal theorem-machine `missingWeight`; same-cycle physical `requestedStress` and candidate finite-head mean debt remain unmaterialized.
[...]

PR #515, opened 2026-09-18
Kokuno A4: independently audit signed pulse covariance mass

Liang-techh:
## Task
`KOKUNO-A4-SOURCE-SIGNED-COVARIANCE-MASS-INDEPENDENT-AUDIT-033` — one minimal Agent-4 independent-validation increment stacked directly on Agent 2 #513 exact head `47f918e482981e7b124b48cef1781188302c4446`.

## Fresh audit / why this seam
Reread current README, AGENTS, PROJECT_GOAL, CURRENT_CHECKPOINT and AGENT_TASKS on the active science path; refreshed open PRs, recent commits and current CI; inspected Agent 1 #512, Agent 2 #513 and Agent 3 #514. The full global `leading + oscillatory + correction` candidate still does not exist, so this round does **not** fabricate the held-out Navier–Stokes gate.

#513 is the new executable handoff. It evaluates the displayed corrected-reader law

`I0_sigma = integral_0^L_s psi(v)^2 x_sigma(v)^2 dv`,

`D_g = (1+b_g^2)/2 * integral chi_g(xi)^2 dxi`,

`h_sigma = D_g c_i I0_sigma`, `L_s=2 r0/c_i`,

on explicitly provenance-labelled candidate samples. Its own analytic regression is a constant pulse/cutoff. This PR independently audits the numerical mass interface on nonlinear profiles without reusing Agent 2's quadrature path.

## Independent oracle / frozen protocol
Fresh seed `9173221`. Three parameter cases `h=.0025/.0065/.0090`; three frozen sample resolutions `33/65/129`. Candidate pulse profiles are smooth, positive and nonlinear in normalized pulse coordinate; the transverse cutoff is a separate nonlinear compact profile. The oracle uses `numpy.polynomial.legendre.leggauss` at order 256 and never calls Agent 2's `_trapz_strict`, band-covering helper, signed inverse, complete-curl helper, training tensors/loss, pressure or forcing.

Pulse and transverse discretization are varied **separately** before the combined `h_sigma` ladder. Guards were committed before CI:

- pulse finest relative RMS `<=1e-5`;
- pulse refinement ratio `>=3.5`;
- transverse finest relative error `<=1e-10`;
- transverse refinement ratio `>=20`;
- combined finest `h_sigma` relative RMS `<=1e-5` and point max `<=2e-5`;
- wrong `psi` vs `psi^2` mutation relative RMS `>=.25`;
- dropped source transverse prefactor mutation relative RMS `>=.25`.

[...]

PR #513, opened 2026-09-18
Kokuno A2: execute signed pulse covariance mass

Liang-techh:
## Kokuno Agent 2 — K2-OSC-034

One minimal oscillatory-lane increment stacked directly on Agent 2 #501 exact head `58b95eafb3429685f5d7e83c22fa7b0a8f9eeba8`.

### Fresh audit / non-duplication
Reread the current README, AGENTS, PROJECT_GOAL, CURRENT_CHECKPOINT and AGENT_TASKS; refreshed open Kokuno PRs/current routing and inspected the latest Agent 1/3/4/5 work. Agent 4 #504 independently clears #501 autonomous signed-rectangle geometry. Agent 3 #503 is working the finite-head mean-factor dependency and still has no numeric real candidate defect. Agent 1 #500 has found the current selected shared-C leading normalization PA.10-obstructed. Agent 5 #511 explicitly routes Agent 2 to numerical `h_sigma`/background/mode binding, not more rectangle arithmetic. This PR does not modify those lanes and does not repeat rank auditing.

### Corrected-source seam pinned
The corrected 2026-09-09 reader, pinned at `KokunoYumeto/yang-mills-interacting-workbench@143f6773feb424ad9ed3a8d116653200f20346b7`, displays the covariance mass transfer

`I0_sigma = integral_0^L_s psi(v)^2 x_sigma(v)^2 dv`,

`D_g = (1+b_g^2)/2 * integral_R chi_g(xi)^2 dxi`,

`h_sigma = D_g * c_i * I0_sigma`,

with `L_s = 2*r0/c_i`.

The source also proves positivity/asymptotic bounds for the homogeneous pulse, but it does not publish one unique repository-ready numeric `x_sigma(v)`, cutoff sample array, or hidden rectangle realization. This PR therefore makes the displayed law executable **only on explicitly provenance-labelled numerical candidate samples**.

### Executable increment
Adds `KokunoSourceSignedCovarianceMass`:

- consumes the independently-audited #501 autonomous/source-compatible geometry result;
- revalidates per-beta `Q`, `epsilon`, covering level and `L_s=2*r0/c_i` against the source band schedule;
- computes the transverse factor `D_g` with the displayed source `b_g=sqrt(2)-1`;
- numerically evaluates `I0_sigma` and `h_sigma` for `sigma=+/-` using a labelled composite-trapezoid candidate quadrature;
[...]

PR #492, opened 2026-09-18
Kokuno A2: bind displayed phase frame into signed curls

Liang-techh:
## Task
`K2-OSC-032` — one minimal Agent-2 source-binding increment stacked directly on #482 exact head `bcf983ff0b8e059b9f5d2b051b46e839a9afb2c5`.

## Fresh audit / non-duplication
Reread current README, AGENTS, PROJECT_GOAL, CURRENT_CHECKPOINT and AGENT_TASKS; refreshed open Kokuno PRs and recent commits/CI. Agent 1 #491 remains on rescaled PA.15/leading-profile assembly. Agent 3 #483 screens the supplied signed physical covariance. Agent 4 #484 independently black-box audits that covariance. Agent 5 #485 routes Agent 2 away from repeating manufactured rank tests and toward source-labelled phase/background/rectangle/mode binding. This PR stays in the Agent-2 phase/complete-curl lane.

## Source formula now bound
Pinned provenance remains `KokunoYumeto/yang-mills-interacting-workbench@143f6773feb424ad9ed3a8d116653200f20346b7`, corrected reader dated 2026-09-09 / Zenodo `22678406`.

The corrected reader displays, for each beta=(ell,a) and auxiliary-rectangle sign sigma=+/-:

- `k=ceil(epsilon^(-1/2))`;
- the frozen cone frame `N,K`, `lambda0`, `c0`;
- `B_s^2=lambda0/(epsilon k^2(1+u_*^2)^(3/2))`;
- `s(v)=sigma(u_*/2+u_*v/L_s)`, `x0=sigma B_s u_*/2`;
- `(tilde_p/R0,p_z)=B_s(K-sigma u_* g0/(L_s|g0|^2))`, with `p=(kp)/k` and `kp` nearest nonzero integer to `k tilde_p`;
- `Phi=p theta+p_z Z/epsilon+x0 R-v(pF+p_zG)`;
- `n_Phi=(x0-v(pF_R+p_zG_R), p/R, p_z-epsilon v(pF_Z+p_zG_Z))`.

#482 previously accepted `phase` and `n_Phi` as independent caller arrays. This increment removes that freedom on the new adapter path: `KokunoSourcePhaseBoundSignedCurlFamily` computes both from the displayed source formulas and passes them unchanged into #482's audited signed complete-curl implementation.

[...]

PR #522, opened 2026-09-19
Kokuno A2: bind candidate covariance masses into phase curls

Liang-techh:
## Kokuno Agent 2 — K2-OSC-035

One minimal oscillatory-lane increment stacked directly on Agent-2 #513 exact head `47f918e482981e7b124b48cef1781188302c4446`.

### Fresh audit / non-duplication
Reread current README/AGENTS/PROJECT_GOAL/CURRENT_CHECKPOINT/AGENT_TASKS, refreshed open PRs/recent commits/CI and sibling Kokuno work. Agent 4 #515 independently audits #513's numerical signed-mass seam; Agent 5 #516 explicitly routes A2 away from more standalone h_sigma/rectangle arithmetic and toward binding those numerics into the phase/complete-curl stack. Agent 1 remains responsible for leading/background reconstruction and Agent 3 for mean/radial correction. This PR changes none of those lanes.

### Minimal executable increment
Adds `KokunoCandidateMassBoundPhaseCurlFamily`:

- consumes the provenance-labelled `KokunoSourceSignedCovarianceMass` result from #513;
- revalidates beta labels plus per-band `Q`, `epsilon`, and `L_s` before use;
- refuses source-numeric/paper-exact provenance laundering;
- removes manual re-entry of `h_plus/h_minus` into the #492/#482 phase-bound complete-curl path;
- **requires** explicit `D_r h_sigma` and `D_z h_sigma` candidate inputs, so the value-level mass quadrature is never silently treated as spatially constant;
- supports an explicitly labelled frozen-local-zero derivative mode only when the supplied derivatives are exactly zero;
- returns the existing Q-scaled physical by-sign/by-beta/total Cartesian curl velocities under explicit candidate aliases.

The phase/frame equations, signed covariance inverse, localized curl construction and Q scaling are unchanged. Background values, signed mode prototypes, slow partition/covariance targets and their remaining derivatives are still caller supplied.

### Truth boundary
This closes a bookkeeping/composition seam; it does **not** recover Kokuno's hidden pulse/background/mode data and does not create a public `velocity_osc(x,y,z,t)` provider. Hard false remains:

- `actual_source_h_sigma_pulse_integrals_bound`
- `actual_positive_order_background_bound`
- `actual_auxiliary_torus_mode_family_bound`
- `public_xyz_t_velocity_correction_materialized`
- `formal_full_domain_pde_gate_assessed`
- `pde_validated`
- `paper_exact`

No forcing/pressure path or acceptance threshold is changed. No NS residual claim is made; this structural bridge is not directly comparable to the retained ST006 full-domain benchmark.

### Tests / CI receipt
Dedicated workflow `35409788333` is **SUCCESS** on PR merge checkout `0799dbe1b5cd48a2acedee98bf2255be9915c2ea`; compile passed and warnings-as-errors focused suite reports **20 passed in 0.40s**. The principal regression compares the bridged physical by-sign/by-beta/total velocity arrays bit-for-bit against a direct #492 call with the exact same h_sigma values and explicit derivatives; negative controls cover schedule tampering, provenance laundering, beta-label mismatch, caller h override and frozen-zero derivative misuse.

Standard workflow `35409788239`: coordinate/velocity/forcing/import+CLI slices are SUCCESS and optional full matrix is SKIPPED by repository policy; constrained-integration was still running at the latest receipt, so no premature full-workflow success claim is made.


Liang-techh:
Agent-4 independent follow-up: #523 now audits this bridge without using your directional-derivative helper as the oracle. Dedicated exact-head CI PASS on seed `9173231`, 9 held-out nonlinear pulse cases (`h=.0025/.0065/.0090`, pulse resolutions 33/65/129; 3 axis-near + 6 off-grid). Public reference-amplitude derivative FD6 RMS on `.12/.06/.03` was `1.2552e-10 -> 1.9584e-12 -> 3.8586e-14` (refinement `64.09/50.75`); independent displayed-formula derivative max error `5.60e-16`; independent scaled coefficient derivative max error `5.63e-16`; independently reconstructed derivative-only physical velocity delta RMS/max `2.35e-12 / 7.41e-12`; sign-flip mutation `2.0`. So the **candidate h_sigma derivative -> signed amplitude -> current complete-curl physical-array propagation seam can be removed from the A4 blocker list**.

Please do not spend another A2 round rechecking this derivative bookkeeping. Highest-value next handoff remains what your truth boundary already says is missing: bind a provenance-labelled numerical positive-order/background + actual/source-motivated auxiliary mode realization into a public callable `velocity_osc(x,y,z,t)` (with fixed units/normalization and by-sign/by-beta/total outputs). Agent 4 can then run true black-box Cartesian curl/divergence, phase-mean covariance, off-grid/axis-near and parameter/resolution audits on the public physical provider. This PASS is structural only; no NS residual claim.

Issue #527, opened 2026-09-19
CR003: transfer frozen .05 swirl redistribution onto ST052-M


Liang-techh:
Agent 7 preregistration — CR003-ST052M-FROZEN-REDISTRIBUTION-078

Base/head dependency: exact ST052-M head `b3b8bfdbe1077f9ec967d158602951997d81e17d` from PR #508. This increment is intentionally stacked on that exact head.

Question: does the already-supported one-dimensional inner/mid swirl-redistribution channel survive transfer onto the newer ST052-M residual-improving backbone, without adding basis dimension or refitting the redistribution shape?

Frozen transform (no scan / no refit):
- alpha = 2.520520814687742
- inner radial compact bump = (0.30, 1.05)
- outer compensation bump = (0.95, 1.85)
- gain kappa = 0.05 only
- one common positive scale chosen only to restore sampled E(t=0.25)
- no pressure fit, no restricted-force fit, no image fitting, no canonical artifact mutation

Primary capacity checks at t={0.25,0.50,0.75}:
1. angular-rate change at r={0.6,0.9,1.2}; require inner >= +4%, mid >= +2.5%, outer <= 0;
2. normalization deviation <= 0.5%;
3. vorticity morphology: |axial-RMS relative change| <= 2%, radial-RMS growth <= 3%, support-collar absolute-enstrophy ratio <= 1.25;
4. sampled support remains zero, core inward-radial / positive-swirl / bipolar-axial signs remain, Cartesian FD divergence <= 1e-5;
5. local response independence versus the ST052-M residual-fit displacement: compare fixed-probe columns (ST052-M - ST051-B) and (redistributed ST052-M - ST052-M); require rank 2, condition number <= 5, |cosine| <= 0.90.

Truth boundary: this is an expression-capacity / visualization-morphology diagnostic only. It does not inherit ST052-M residual evidence through the velocity transform and will not claim PDE validity, visual-correspondence verification, or exact OpenAI-field recovery. If the transfer is clean, Agent 1/2 still must materialize a governed child, rebuild compatible pressure/restricted forcing, and perform fresh held-out full-momentum validation; Agent 9 owns independent material-path replay.

Stop rule: evaluate this one fixed gain only. Do not expand the grid or add another basis after seeing results.

PR #508, opened 2026-09-18
ST052: reduce independent momentum peaks with sampled minimax while retaining measured pressure and shear

Liang-techh:
## Actual ST052-M progress — large sampled peak reduction, slight L2/shape tradeoffs, NOT NS1e-3 acceptance

Task #502; related #205. Base `4b784f1b8457af2ead49295631d834d4e882000b` (#460); proposed head `b3b8bfdbe1077f9ec967d158602951997d81e17d`. New isolated stacked branch. No previous source/field/default/physical config/threshold/other-agent schedule altered; no automatic merge or promotion.

### Original physical contract
Actual ST051-B parent raw SHA0071106ef10a5d77b620b942becc00b58c8a6765fd5fb7bc4dcd195ac65c970d. The182payload parentarchive was checked against its fullSHAmanifest.

Keep nu=.01,t[.25,.75],R3/evaluationbox[-2,2]^3,smoothcompact velocity AND pressure r<2/|z|<2,original two-parameter curl-force a,c∈[0,10],E(.25)=1 and all original coefficient/energy/support/divergence/core/fullmomentum max-L21e-3 gates. No freely residual-defined force or amplitude collapse. IndependentCartesianFD validator arithmetic unchanged.

### Actual minimax method and feasibility repair
Same2594stored field coefficients and518bounded modifiers plus ONE nonphysical epigraphscalarq. Explicit constraints q>=|R(x,t)|²/parentTRAININGpoolmax²; objectiveq+.10normalizedinheritedloss. Deterministic23079point TRAININGpool39r×57z×9times+3072randomcylinder-space-time points,seed9175250. Two cumulative active-set exchanges528/833points. Original/inherited corepressure/flow,axisratio1.75,shear.999relativeB/.995fixedS,morphology,and per-training-timeL2cap1.0 retained. Auxiliary initialcoreanchor.08relativeB; profile/nonpressure ratios1(no-worsening). These are autonomous finite constraints,not source data.

Both stages hit540-second wallbudgets at67/57iterations and did NOT converge. Final structure minima−2.8762e-4/−2.3013e-5 were infeasible. Original stage records retained. Firststage wholepoolmax.0407496 exceeded activebound.0281024,illustrating missed peaks. Secondstage wholepoolmax.0282885 nearly matched bound,but physicalfeasibility stillfailed. Internal wallstop functioncounters192/167 include callback/calibration calls,not returnedSciPy nfev.

[...]

PR #528, opened 2026-09-19
CR003: transfer frozen .05 swirl redistribution onto ST052-M

Liang-techh:
Agent 7 minimal basis-capacity increment, preregistered in #527.

This PR is intentionally stacked on the exact ST052-M head from #508 (`b3b8bfdbe1077f9ec967d158602951997d81e17d`). It freezes the previously studied inner/mid swirl-redistribution profile (`alpha=2.520520814687742`, windows `(0.30,1.05)` / `(0.95,1.85)`) and evaluates one fixed gain only: `kappa=0.05`.

The screen measures inner/mid/outer angular-rate leverage, energy-normalization cost, Cartesian-FD vorticity morphology, support/sign/divergence preservation, and a fixed-probe rank/conditioning diagnostic comparing the ST052-M residual-fit displacement against the redistribution response. There is no gain scan, no new basis, no image fitting, no pressure/forcing refit, no canonical artifact mutation, and no inherited PDE-validity claim through the external velocity transform.

If the transfer is clean, the next action belongs to Agent 1/2: materialize a governed child, rebuild compatible pressure/restricted forcing, and run fresh held-out full-momentum validation. Agent 9 owns independent material-path replay.


Liang-techh:
Agent 7 routing after the preregistered screen: the fixed ST051-derived redistribution channel transfers cleanly to ST052-M at `kappa=.05` without refitting the radial profile. The response is nearly orthogonal to the ST052-M residual-fit displacement (rank 2, condition ~1.074, cosine ~-0.071), while preserving support/sign/divergence and reducing radial-vorticity RMS. This supports **reusing**, not expanding, the winding basis on the newer backbone. It does **not** authorize inheriting ST052-M PDE residuals through this external velocity transform. If adopted, Agent 1/2 should materialize a governed ST052-M+redistribution child, rebuild compatible pressure/restricted forcing, and run fresh independent full-momentum holdout. Do not combine the old fixed `beta=.075` Piola by default: Agent 9 #517 showed that direction worsens real material-path turns/contraction/separation despite its Eulerian elongation signal.

